PR #91, opened 2015-05-22
Authz tunning part 2

tpazderka:
uic dict is now trimmed in id_token as well


Zelvuska:
Hi Roland,

Would you mind merging this or do you have some objections against this?

Thanks,
Tomas

PR #107, opened 2015-07-23
Removed dependency on dict-like behaviour from SessionDB and Provider

tpazderka:


ziima:
That's a pretty small change :)

I miss the alternative `SessionDB.db` implementation which you suggested. That would allow tests to ensure the tests are not dependent on using `dict` as a database.

Required API of `SessionDB.db` should be documented now when you still remember which it is.

PR #111, opened 2015-08-17
Disable refresh token by default

tpazderka:
As a mean to access user data without presence, it should be only issued
when offline_access is requested and allowed by user.

### DIFF
--- a/src/oic/oauth2/provider.py
+++ b/src/oic/oauth2/provider.py
@@ -776,7 +776,7 @@ class Provider(object):
             assert areq["redirect_uri"] == _info["redirect_uri"]
 
         try:
-            _tinfo = _sdb.upgrade_to_token(areq["code"])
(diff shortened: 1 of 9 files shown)


tpazderka:
Rebased on current master

rohe:
Applied

tpazderka:
Dear Roland,

This doesn't look merged, is that a mistake or intentional?

Thanks

rohe:
Not intentional. Couldn't just merge it since the master had changed since you issued the request.
Must have done something wrong then.

rohe:
Had to reverse this pull request it was not compatible with the OP test tool. Have to study it more thoroughly before I can apply it.

tpazderka:
Could you, please, provide more info on what is not compatible? I can redo this branch to comply with your requirements.

PR #114, opened 2015-08-20
Endpoint refactor

tpazderka:
Endpoint class simplified.

Removed unused methods and added url parameter.
Endpoint matching is done by etype and url construction by url parameter. This allows overrides on url creation.


tpazderka:
Rebased on current master

PR #118, opened 2015-08-24
Updated sub creation

tpazderka:
During client registration a random salt is generated and stored in
`cdb`. This is used in pairwise sub creation.
AuthnEvent now takes user-specific salt which can be returned from
authenticated_as() and is used in public and pairwise sub creation.


tpazderka:
Rebased on current master

PR #130, opened 2015-09-28
Access Token Response token_type value is case insensitive

zzgvh:
According to section 4.2.2 of [RFC 6749](https://tools.ietf.org/html/rfc6749#section-4.2.2) the token_type
value is case insensitive.

In two places in oic/oic/\_\_init\_\_.py this value is checked against
using the string "Bearer" resulting in an exception when the auth
server response token_type value is "bearer".

Fix by applying lower() to the token_type value and compare with
"bearer".

We experienced this when experimenting with [Keycloak](http://keycloak.jboss.org/) for SSO.

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -702,7 +702,7 @@ class Client(oauth2.Client):
 
             if token.is_valid():
                 uir["access_token"] = token.access_token
-                if token.token_type == "Bearer" and method == "GET":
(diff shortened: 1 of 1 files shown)


rohe:
A test failed so I did this by hand.
Added your changes and fixed the test all in a update attributed to you.

zzgvh:
Ah, thanks a bunch!

PR #138, opened 2015-10-23
UserInfo token error

tpazderka:
Return error response if token is missing or malformed on userinfo
endpoint


ziima:
Test is missing.

tpazderka:
Test for endpoint added.

PR #140, opened 2015-10-26
Refresh token disabled by default

tpazderka:
As a mean to access user data without his presence, it should be only issued
when offline_access is requested and allowed by user.

This is another attempt for a closed #111, I found a bug in my implementation. This should be working now.

### DIFF
--- a/src/oic/oauth2/provider.py
+++ b/src/oic/oauth2/provider.py
@@ -782,7 +782,7 @@ class Provider(object):
             assert areq["redirect_uri"] == _info["redirect_uri"]
 
         try:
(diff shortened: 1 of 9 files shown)


tpazderka:
This looks like it was merged. So I am closing it.

rohe:
> 2 nov 2015 kl. 08:03 skrev tpazderka <notifications@github.com>:
> 
> This looks like it was merged. So I am closing it.

Yeah, had to do it manually !

— Roland
'Look, that's why there's rules, understand? So that you think before you break ’em.’ - Terry Pratchett

PR #152, opened 2015-12-18
Produce ErrorResponse if scopes missing completely

tpazderka:
OIC verification fails with KeyError if scopes are missing completely.

It should instead produce an error message.

### DIFF
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -673,6 +673,12 @@ class Message(object):
     def __getitem__(self, item):
         return self._dict[item]
 
+    def get(self, item, default=None):
+        try:
+            return self[item]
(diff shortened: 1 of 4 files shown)


tpazderka:
I am not sure why the python 3.5 tests fail, they do pass locally (both in py3.4 and 3.5)...

rohe:
> 18 dec 2015 kl. 14:15 skrev tpazderka <notifications@github.com>:
> 
> I am not sure why the python 3.5 tests fail, they do pass locally (both in py3.4 and 3.5)…

Yeah, I was surprised about that too.
Specially since nothing in your pull request modifies/touches what’s in the failed tests.
Haven’t had time to look at it more thoroughly yet.

— Roland

PR #156, opened 2016-01-11
Don't do utc time for 'iat' claim in ID Token.

rebeckag:


rohe:
> 11 jan 2016 kl. 16:32 skrev Rebecka Gulliksson <notifications@github.com>:
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/rohe/pyoidc/pull/156
> 
> Commit Summary
> 
> 	• Don’t do utc time for 'iat' claim in ID Token.

RFC 7519 says you MUST use NumericDate for ’exp’, ’nbf’ and ’iat’.
And Numericdate is defined to be:

'A JSON numeric value representing the number of seconds from
1970-01-01T00:00:00Z UTC until the specified UTC date/time,
ignoring leap seconds.’

Hence, UTC it is.

rebeckag:
The previous implementation of time_util.utc_time_sans_frac, using time.mktime which expects local time, was not returning UTC timestamps. Should be correct now!

PR #169, opened 2016-02-11
Remove state hash

rebeckag:


rohe:
Replaced with modifications that follow the present proposal which is using state instead of state_hash.

PR #180, opened 2016-03-09
Fixed the links and a bullet error

danielquinn:
This may be what is preventing updates to readthedocs.


rohe:
No, the automatic github modification -> readthedocs update had for some reason just been removed.
I've added it now.

Issue #186, opened 2016-03-24
Unicode Cookies Break CherryPy


serac:
It appears that b1953a1d7cae2168312f7a76ec440258c8d08dc3 and related commits create unicode headers, which appears to break cherrypy on Python 2.7:

```
WSGI response header key u'Set-Cookie' is not of type str.
['Traceback (most recent call last):\n', '  File "/Users/marvin/Documents/Develop/pyoidc-op/src/server.py", line 374, in application\n    return callback(environ, start_response, logger)\n', '  File "/Users/marvin/Code/pyoidc/src/oic/oauth2/provider.py", line 66, in __call__\n    return self.func(*args, **kwargs)\n', '  File "/Users/marvin/Documents/Develop/pyoidc-op/src/server.py", line 172, in authorization\n    logger=logger)\n', '  File "/Users/marvin/Code/pyoidc/src/oic/utils/http_util.py", line 380, in wsgi_wrapper\n    return resp(environ, start_response)\n', '  File "/Users/marvin/Code/pyoidc/src/oic/utils/http_util.py", line 114, in __call__\n    start_response(self.status, self.headers)\n', '  File "/Library/Python/2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 2173, in start_response\n    raise TypeError("WSGI response header key %r is not of type str." % k)\n', "TypeError: WSGI response header key u'Set-Cookie' is not of type str.\n"]
AssertionError('WSGI start_response called a second time with no exc_info.',)
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 1302, in communicate
    req.respond()
  File "/Library/Python/2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 831, in respond
    self.server.gateway(self).respond()
  File "/Library/Python/2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 2135, in respond
    response = self.req.server.wsgi_app(self.env, self.start_response)
  File "/Users/marvin/Documents/Develop/pyoidc-op/src/server.py", line 381, in application
    return resp(environ, start_response)
  File "/Users/marvin/Code/pyoidc/src/oic/utils/http_util.py", line 49, in __call__
    start_response(self.status, self.headers)
  File "/Library/Python/2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 2157, in start_response
[...]

PR #187, opened 2016-03-24
Issue #186 Force str type for HTTP header fields.

serac:
Candidate fix for #186.

### DIFF
--- a/src/oic/utils/http_util.py
+++ b/src/oic/utils/http_util.py
@@ -97,7 +97,7 @@ class Redirect(Response):
 
     def __call__(self, environ, start_response, **kwargs):
         location = self.message
-        self.headers.append(('location', location))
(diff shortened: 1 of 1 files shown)


rohe:
Well, this breaks a number of tests so I can't let it in.
We'll have to fix that first.

serac:
Understood. Do you agree with the premise that header field names/values ought to be ASCII strings?

rohe:
Well, by default message (RFC 2616, 5987) header fields in HTTP can not carry characters outside the ISO 8859-1 character set. So yes !

serac:
Ok, that was my analysis as well.

lwm:
@serac, any update on this? Would be nice to get this merged soon!

lwm:
Well, that says it all :smile:

PR #213, opened 2016-07-13
Convert 'bearer' to 'Bearer' for HTTP headers

koriaf:
Some software (like google) has strict requirements for it, and specification says `should` about `Bearer`. So converting any lowercase header here to capitalized. Headers like JWT or Someheader shall be untouched.

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -759,7 +759,7 @@ class Client(oauth2.Client):
             else:
                 # use_authorization_header, token_in_message_body
                 if "use_authorization_header" in _behav:
-                    token_header = "{type} {token}".format(type=_ttype,
+                    token_header = "{type} {token}".format(type=_ttype.capitalize(),
                                                            token=_token)
                     if "headers" in kwargs:
                         kwargs["headers"].update({"Authorization": token_header})


src/oic/oic/__init__.py
```diff
@@ -759,6 +759,9 @@ def user_info_request(self, method="GET", state="", scope="", **kwargs):
             else:
                 # use_authorization_header, token_in_message_body
                 if "use_authorization_header" in _behav:
+                    if _ttype.lower() == _ttype:
+                        # fix 'bearer' to 'Bearer' (standard requirement)
+                        _ttype = _ttype.capitalize()
                     token_header = "{type} {token}".format(type=_ttype,
```
lwm: Can't we just call `_ttype.capitalize()` here and forget the `if` check?

```
>>> "foo".capitalize()
'Foo'
>>> "Foo".capitalize()
'Foo'
```

koriaf: Yes, in this case it works of course.
My initial idea was to add a warning here and hunt out all source places where this problem is coming from, but I haven't had time and have forgotten about it.
Not sure we aren't having any other problems because lowercase word is being sent from somewhere else.

rohe: I think we should go by the old Postel 'rule': be liberal in what you accept from others.
So, 'bearer', 'BEARER', 'Bearer'  and so on all should be accepted.

koriaf: But in this code we do not accept it, we send it. Please note RFC says about it: https://tools.ietf.org/html/rfc6750#section-2.1

rohe: You're correct my bad. We MUST send 'Bearer'.

rohe: I'll go for Luke's proposal: just capitalize.

koriaf: Fixed in place, thanks all for the proposal.

lwm:
:+1:

Issue #221, opened 2016-08-09
After timeout, remote key is removed in case of 304 response


fmoggia:
In keyio.py _uptodate checks the timeout for caching the key and if the timeout expired invokes update() to force a reload of the key.

```
    def _uptodate(self):
        res = False
        if self._keys is not []:
            if self.remote:  # verify that it's not to old
                if time.time() > self.time_out:
                    if self.update():
                        res = True
```

The update method reset self._keys to an empty array and invokes do_remote()

```
            self._keys = []

            if self.remote is False:
                if self.fileformat == "jwk":
                    self.do_local_jwk(self.source)
                elif self.fileformat == "der":
                    self.do_local_der(self.source, self.keytype, self.keyusage)
            else:
                res = self.do_remote()
```

do_remote() contact the jwks url to retrieve the key and receives back a 304 Not Modified and doesn't set back the cached key into self._keys

```
 if r.status_code == 304:  # file has not changed
            self.time_out = time.time() + self.cache_time
            self.last_updated = time.time()
            return False
```

This is causing verification of JWT tokens to fail after the default timeout expires, with the following error:
NoSuitableSigningKeys at /oauth/callback
No key for algorithm: RS256

Shouldn't the handling of the 304 reset the key into the self._keys array  with something like the code below ?

```
        if r.status_code == 304:  # file has not changed
            self.time_out = time.time() + self.cache_time
            self.last_updated = time.time()
            try:
                self.do_keys(self.imp_jwks["keys"])
            except KeyError:
                logger.error("No 'keys' keyword in JWKS")
                raise UpdateFailed(
                    "Remote key update from '{}' failed.".format(self.source))
            return False
```

fmoggia:
Looks like the issue was addressed in 2c7239373ba1bf3185b48faa6b14b12ff19502c9

PR #222, opened 2016-08-09
After timeout, remote key is removed in case of 304 response #221:

fmoggia:
- Added logic that fills in the existing keys in case of a 304 response from the keys source.


rohe:
Accepted the gist of it.

PR #230, opened 2016-10-13
Sanitizes log entries to remove sensitive data

### LINKED ISSUE #228: Security: Logging captures credentials
Some of the `debug` logging e.g. https://github.com/rohe/pyoidc/blob/master/src/oic/utils/authn/user.py#L306 captures raw cleartext passwords as entered by users. 

And even the `info` loglevel captures access tokens in the responses: https://github.com/rohe/pyoidc/blob/master/src/oic/oic/provider.py#L988

This is potentially dangerous, and I think it would be better to sanitize any credentials or secrets before logging.

bjmc:
Fixes #228

- Adds `oic.utils.sanitize.sanitize()` function that makes a best-effort attempt to remove sensitive data from mappings and strings. 
- Applies `sanitize()` throughout the codebase on log entries.
- Removes or modifies some log entries I wasn't confident could be sanitized, or that seemed like they would be uninformative after sanitizing. 
- Makes `oic.oauth2.message.Message` into a subclass of `MutableMapping` to make life easier for the `sanitize()` function.
- Adds a nicer `__repr__()` for the `KeyJar` to make testing easier.
- Slightly changes sorting behavior on `authn_context` to get consistent test output.
- Adds tests for `sanitize()`
- Extends some existing tests to verify their logs are clean.
- Modifies tox.ini to allow passing extra arguments in to the test runner.

### DIFF
--- a/src/oic/extension/client.py
+++ b/src/oic/extension/client.py
@@ -20,6 +20,7 @@ from oic.oic import OIDCONF_PATTERN
 from oic.oic.message import AuthorizationResponse
 from oic.utils.http_util import SUCCESSFUL
 from oic.utils.keyio import KeyJar
+from oic.utils.sanitize import sanitize
 
 from oic.oauth2.message import ErrorResponse
 from oic.oauth2.exception import Unsupported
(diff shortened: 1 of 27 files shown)


rohe:
I really like this, but in the test tools I write I need to be able to turn of sanitising.

rohe:
I'll accept it and then I'll make sure I can get at the data I need for the sake of the test tools.
Sanitize should be the default.

bjmc:
Thanks! You could easily modify `sanitize()` to check for an environment variable and return unredacted values when running in the context of the test tools.

rohe:
> 17 okt. 2016 kl. 11:11 skrev Brendan McCollam <notifications@github.com>:
> 
> Thanks!

It’s me that should thank you! 
Great work. Have had something similar on my list now for a while but you beat me to it :-)

> You could easily modify sanitize() to check for an environment variable and return unredacted values when running in the context of the test tools.

Something like that, yes !

-- Roland
"Education is the path from cocky ignorance to miserable uncertainty.” - Mark Twain

PR #235, opened 2016-10-28
client_secret_basic authn: include client_id in authz code flow

jgehrcke:
@gpaul discovered that the client_secret_basic authentication method does not comply with RFC 6749, which states in section 3.2.1:

```
In the "authorization_code" "grant_type" request to the token endpoint,
an unauthenticated client MUST send its "client_id"
```

In the current code, the `client_id` may become deleted even after being manually injected:

```
        if cis and not cis.c_param["client_id"][VREQUIRED]:
            try:
                del cis["client_id"]
            except KeyError:
                pass
```

(https://github.com/rohe/pyoidc/blob/eb2daf41e76511ab5b3d92b6d6043d973913c805/src/oic/utils/authn/client.py#L122)

Effectively, the following usage of pyoidc will lead to a non standards-compliant access token request:

```
args = {
    "code": aresp["code"],
    "redirect_uri": client.redirect_uris[0],
    "client_id": client.client_id,
}

resp = client.do_access_token_request(scope="openid",
                                      state=aresp["state"],
                                      request_args=args,
                                      authn_method="client_secret_basic"
                                      )
```

Note that `client_id` was manually specified in `args`, but the key deletion code shown above will still strip it before sending the request.

The request shown above was confirmed to fail against Azure Active Directory's OIDC provider, with the error message pointing towards the missing `client_id` parameter.

I do not precisely understand the `cis.c_param["client_id"][VREQUIRED]` mechanism, so I did not want to use this for tackling the issue. Instead, I took an explicit approach that checks for the two conditions referred to in the quoted RFC section:
- access token request
- authorization_code grant_type

If both conditions are fulfilled, the `client_id` is injected as request parameter.

Can we add a test for this to the pyoidc project?

@rohe I'd appreciate your review, let me know what you think.

### DIFF
--- a/src/oic/utils/authn/client.py
+++ b/src/oic/utils/authn/client.py
@@ -13,7 +13,7 @@ from oic.exception import UnknownAssertionType
 from oic.exception import NotForMe
 from oic import rndstr
 from oic.oauth2 import VREQUIRED
-from oic.oauth2 import AccessTokenRequest
+from oic.oauth2 import AccessTokenRequest, RefreshAccessTokenRequest
 from oic.oauth2 import SINGLE_OPTIONAL_STRING
 from oic.oic import REQUEST2ENDPOINT
 from oic.oic import DEF_SIGN_ALG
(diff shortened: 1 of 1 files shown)


jgehrcke:
Interesting CI result for test `TestClientSecretBasic.test_does_not_remove_padding`
```
            if isinstance(cis, (AccessTokenRequest, RefreshAccessTokenRequest)):
                if "grant_type" in cis:
                    if cis["grant_type"] == "authorization_code":
>                       cis["client_id"] = self.cli.client_id
E                       AttributeError: 'NoneType' object has no attribute 'client_id'
```

So it looks like we can get into a state where `self.cli` is not yet set. I'll add another check for that to prevent the AttributeError, but it looks like this should _never_ be the case if not used in tests.

jgehrcke:
@rohe would appreciate if you could review it, we confirmed this to fix the compatibility between pyoidc on SP side with Microsoft Azure Active Directory on OP side.

rohe:
Will do, a bit short on time right now.
Hopefully tomorrow at the latest.

jgehrcke:
Thank you so much, appreciated! And if it does not happen, then it does not happen.. :-) that's the open source game.

rohe:
Just uploaded a fix to 235, so I'm closing this.

jgehrcke:
Thanks!

PR #238, opened 2016-11-08
Simplify example of token request in client cookbook.

zamzterz:
pyoidc is clever enough to fill in the blanks for
client_id, client_secret, redirect_uri, etc. based on the state value.


lwm:
Thanks @zamzterz! Sorry it took so long to merge this :|

PR #256, opened 2017-01-18
Drop the unmaintained examples.

lwm:
Follow on from https://github.com/rohe/pyoidc/issues/254#issuecomment-273405791:

  * Delete oidc_fed/, oauth_example/ and unused op/rp examples in oidc_example
  * Rename oidc_example -> examples
  * Rename op2 -> op, and rp3 -> rp
  * Merge docker code from its-dirg/docker with op/rp in examples/

Another huge diff; Sorry. You can review [the resulting structure here](https://github.com/rohe/pyoidc/tree/57e1ed56a00bbc6fd307f0af1fb8b7416807ed19/examples).

I would imagine this is the last of the large diffs. The rest will be fixups on the examples (which I will follow up with in any new PRs). You'll most likely need to update the docker images, but it can be done once the code runs ;)

If I removed anything you wanted, please let me know and I can fix it up.


rohe:
Not OK with removing oidc_fed.

lwm:
> Not OK with removing oidc_fed.

Dropped that commit!

lwm:
Failure is unrelated to these changes. Possibly related to https://github.com/rohe/pyoidc/commit/88d21c318390a55efd8c099d124c5d0945c83632?

rohe:
I hate it when I can't replicate the problems Travis sees on my machine.
The Python2 I'm using is 2.7.6 and Travis uses 2.7.9 which my be one source.

lwm:
> The Python2 I'm using is 2.7.6 and Travis uses 2.7.9 which my be one source.

It could be. I absolutely recommend [pyenv](https://github.com/yyuu/pyenv) for this kind of fast switching of Python versions. Takes the pain out of it, all installations are isolated!

I think I saw that test fail before. I could also recommend [pytest-repeat](https://pypi.python.org/pypi/pytest-repeat), in the case it is 'flaky' and you can just run it 100 times until you see the failure.

lwm:
Yes, it's definitely flaky. It fails only some times and it is related to the test logic.

PR #272, opened 2017-03-02
allow for a "dummy" in webfinger response

zandbelt:
for the RP test rp-discovery-webfinger-unknown-member


rohe:
Fixed some of this in #273

PR #274, opened 2017-03-02
All the OIDC federation code is moved into a repository of its own.

rohe:
https://github.com/OpenIDC/fedoidc


lwm:
# −11,435

:beers:

PR #279, opened 2017-03-03
Add pip-tools for dependency management.

lwm:
So, this PR adds the following two Make targets:

```
    $ make install  # install all python dependencies
    $ make upgrade  # upgrade all python dependencies
```
Please see http://nvie.com/posts/pin-your-packages/ and https://github.com/jazzband/pip-tools to get up to speed on why this important and what `pip-tools` does.


rohe:
I like this !

lwm:
:)

PR #281, opened 2017-03-03
fix oic federation reference

zandbelt:
remove "oic/federation" from packages provided in setup.py


rohe:
Two minds with but a single thought :-)

Issue #284, opened 2017-03-13
Bad Code causes KeyError instead of TokenErrorResponse


tpazderka:
Comparison in `token_endpoint` [here](https://github.com/OpenIDC/pyoidc/blob/master/src/oic/oic/provider.py#L1036) does not consider the eventuality that the code could be missing from `SessionDB`. This can happen for multitude of reasons and this should result in `TokenErrorResponse` rather than uncaught `KeyError`.

The following test demonstrates the issue.
```python
    def test_token_endpoint_bad_code(self):
        authreq = AuthorizationRequest(state="state",
                                       redirect_uri="http://example.com/authz",
                                       client_id=CLIENT_ID,
                                       response_type="code",
                                       scope=["openid"])

        _sdb = self.provider.sdb
        sid = _sdb.access_token.key(user="sub", areq=authreq)
        access_grant = _sdb.access_token(sid=sid)
        ae = AuthnEvent("user", "salt")
        _sdb[sid] = {
            "oauth_state": "authz",
            "authn_event": ae,
            "authzreq": authreq.to_json(),
            "client_id": CLIENT_ID,
            "code": access_grant,
            "code_used": False,
            "scope": ["openid"],
            "state": "state",
            "redirect_uri": "http://example.com/authz",
        }
        _sdb.do_sub(sid, "client_salt")

        # Construct Access token request
        areq = AccessTokenRequest(code='bad_code',
                                  client_id=CLIENT_ID,
                                  redirect_uri="http://example.com/authz",
                                  client_secret=CLIENT_SECRET,
                                  grant_type='authorization_code',
                                  state="state")

        txt = areq.to_urlencoded()

        resp = self.provider.token_endpoint(request=txt)
        atr = TokenErrorResponse().deserialize(resp.message, "json")
        assert atr['error'] == "access_denied"
```

tpazderka:
Added. PR is failing on the right place :)

PR #285, opened 2017-03-13
Failing test for Issue #284

tpazderka:
Test to confirm #284 

I am not sure if the `assert` for error should be `access_denied` or `bad_code`. Leaving that up to you...


lwm:
Thanks! Will get to this ASAP.

PR #288, opened 2017-03-20
Fix #283, make the simple_op example to work by fixing the dict iteration

### LINKED ISSUE #283: oauth2 provider redirect_uris iteration
I tried to get `oidc_example/simple_op` up and running, and ran into one issue. When verifying the redirect URI, the provider reported some error:
```
ValueError: need more than 1 value to unpack
```

I checked out the code, and found this piece of code when verifying redirect URI in 
https://github.com/OpenIDC/pyoidc/blob/master/src/oic/oauth2/provider.py#L231-L232:
```
for regbase, rquery in self.cdb[str(areq["client_id"])][
                "redirect_uris"]:
```

It seems `self.cdb[str(areq["client_id"])]["redirect_uris"]` is a dictionary and we are trying to iterate this dictionary and unpack the dictionary item into key (`regbase`) and value (`rquery`).  However, we have to call `items()` [when iterating a dictionary](http://stackoverflow.com/questions/3294889/iterating-over-dictionaries-using-for-loops-in-python), otherwise, we will run into issue like above.

I am not sure if I miss something but I have to modify the source code to append `items()` to get it fixed:
```
for regbase, rquery in self.cdb[str(areq["client_id"])][
                "redirect_uris"].items():
```

Could you confirm if it is a bug or there is something I got wrong? Thanks.

I am using [Python 2.7.13 in Docker's official Python image](https://hub.docker.com/_/python/).

niyue:
As I mentioned in #283, oidc_example/simple_op cannot work because of a dictionary iteration issue. 

In https://github.com/OpenIDC/pyoidc/blob/master/src/oic/oauth2/provider.py#L231-L232, `self.cdb[str(areq["client_id"])]["redirect_uris"]` is a dictionary and we are trying to iterate this dictionary and unpack the dictionary item into key (regbase) and value (rquery). However, we have to call `items()` when iterating a dictionary this way.

I can make the example to work after fixing this issue in the source. (But I made some additional change to the oidc_example/simple_op example itself, which probably can be submitted as another PR later).

### DIFF
--- a/src/oic/oauth2/provider.py
+++ b/src/oic/oauth2/provider.py
@@ -229,7 +229,7 @@ class Provider(object):
 
             match = False
             for regbase, rquery in self.cdb[str(areq["client_id"])][
-                "redirect_uris"]:
+                "redirect_uris"].items():
                 # The URI MUST exactly match one of the Redirection URI
                 if _base == regbase:
                     # every registered query component must exist in the


lwm:
Hmmmm, that change seemed to break a lot of test :)

AFAIK, `self.cdb[str(areq["client_id"])]["redirect_uris"]` comes directly from the client configuration, which in my setup, looks like this:

```
    CLIENT = {
        'frontend-id': { 'redirect_uris': pack_redirect_uri(["http://www.frontend.com"])}
    }
```

Where `pack_redirect_uri` does the following:

```python
>>> pack_redirect_uri(["foo", "bar", "baz"])
[['foo', None], ['bar', None], ['baz', None]]
```

So, I'm not sure this is the right change?

niyue:
Hmm, so the `self.cdb[str(areq["client_id"])]["redirect_uris"]` is not a dictionary but a list of list, where each list in list is of two elements? 

Okay, please reject this PR, and I will take a look if this can be fixed by configuring the simple_op to use this kind of data structure as redirect_uris instead. Thanks.

lwm:
Sure! No problem, we've seen big demand for a working example and we need all the help we can get!

PR #293, opened 2017-03-22
Remove examples. Point to live use cases.

lwm:

---

Since they are unmaintained and I think we would rather focus on improving the documentation and the core library API (in this repository!), I would remove this to reduce the maintenance burden. I've update the README to point users to live examples, so there are still examples out there.


lwm:
Uhhhhh ... I have no idea what happened here haha.

Can we remove the example or no @rohe?

rohe:
I'm not completely convinced that we should remove all examples.
In the best of worlds we should have one well documented example.
Non working examples are worse then no examples.

rohe:
Also not sure what happened.
Seems to a number of changes I have in my branch but I never sent a pull request to my knowledge ??

lwm:
> Non working examples are worse then no examples.

Sounds fine to me then!

> Seems to be a number of changes I have in my branch

I must have pulled from there instead of the branch on this repository :smile:

Issue #287, opened 2017-03-20
Fix flaky test


lwm:
Didn't take much time on this but first seen in https://travis-ci.org/OpenIDC/pyoidc/jobs/213191415.

tpazderka:
I was digging a little bit into this.

The issue is in [sanitize](https://github.com/OpenIDC/pyoidc/blob/master/src/oic/utils/sanitize.py#L10) as it will keep `B` or `U` before the `<REDACTED>` if the sensitive part is starting with it.

So it is not an error in test but rather a bug in `sanitize` function...

`sanitize('code=bttt')` => `code=b<REDACTED>`

PR #294, opened 2017-03-22
Fixup documentation building

lwm:

---

So, the two main things I have done here are:

  1) Remove code index generation
  2) Add a separate documentation dependency

Why? For 1), I think the index generation depends on the library being installed, which we cannot have installed on the readthedocs platform because the `pyldap` dependency requires some system dependencies, and readthedocs cannot install anything on the system (phew!).

For 2), well, I don't know why this was missing but it should be a seperate dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,6 @@ foo.*
 
 # Dynamically created doc folders
 doc/_build
-doc/index/*
 !tests/data/keys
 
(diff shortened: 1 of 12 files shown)


lwm:
Failure is unrelated and due to https://github.com/OpenIDC/pyoidc/issues/287.

PR #295, opened 2017-03-23
Wrong token was used to find the session id.

rohe:

- [#295 ] Fixed a bug in the _refresh_access_token_endpoint method that prevented new Access Token from being issued . The second fix was for a spelling error in the verify_redirect_uris method.
The effect of the spelling error was that an exception was thrown if a redirect_uri that a native app tried to register didn't adhere to what the standard says. Added tests for both these fixes.

---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -1175,7 +1175,7 @@ class Provider(AProvider):
     # noinspection PyUnusedLocal
     def userinfo_endpoint(self, request="", **kwargs):
         """
-        :param request: The request in a string format
+        :param request: The request in a string format or as a dictionary
         """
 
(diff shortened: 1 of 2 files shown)


lwm:
Could do with a test and a change log entry. Also, which bug motivated this fix?

rohe:
Doing a token refresh using a refresh token just didn't work.

rohe:
The second one comes into effect when a native client tries to register a redirect_uri that doesn't comply with what the OIDC standard says.

rohe:
OK, I'll add some tests.

rohe:
That Travis fails has nothing to do with the fixes it is the old intermittent error.

rohe:
So what's holding up the acceptance of this ?
Or should I just merge it :-)

lwm:
LGTM!

PR #299, opened 2017-03-27
297 isort

### LINKED ISSUE #297: Using static code control in CI - isort
It might be a good practice to use an static checks of the code. One of the first (and probably easiest steps) would be to add an automated sorting of imports (and incorporate such checks in CI to enforce them).

One of the best tools available is [isort](https://pypi.python.org/pypi/isort) which allows for sorting as well as for checking.
I would also recommend to convert to a more compact style of imports (`2 - Hanging Indent` perhaps).

If you agree then I can take care of this.

tpazderka:

Close #297 
---
None of the above seems necessary.

Two additional targets in Makefile: `isort` - which will sort the imports and `check-isort` used for CI verification.


lwm:
> None of the above seems necessary.

FYI the [PULL_REQUEST_TEMPLATE](https://github.com/OpenIDC/pyoidc/blob/master/.github/PULL_REQUEST_TEMPLATE.md) has a 'ignore this if it doesn't make sense' clause ;)

.travis.yml
```diff
@@ -29,6 +31,8 @@ install:
         pip install -U tox ;;
       docs)
         pip install -U sphinx ;;
+      quality)
+        pip install -rrequirements/quality.txt
```
lwm: I was hoping we could keep this all under the control of Tox. However, I just realise we hard code:

```
[testenv]
deps = -rrequirements/test.txt
```

Which is actually wrong because, for example, we have a `docs.in/txt` for the `docs` case.

I'm not sure how to do this though. So, this is more a note if you're interested in looking into it.

lwm:
Looks good then :+1:. Feel free to ignore above comment.

.travis.yml
tpazderka: Well, we could specify different `testenv` for `docs` and `quality`.

But I would rather do it in separate ticket.

lwm: Sounds good, I've created https://github.com/OpenIDC/pyoidc/issues/301 to keep track of that then.

lwm:
:tada:

PR #304, opened 2017-03-27
Reduce duplication for Tox testing.

lwm:
It reads a bit better.


lwm:
Unrelated failure (good 'ol https://github.com/OpenIDC/pyoidc/issues/287 again).

Issue #298, opened 2017-03-26
Use pylama for code quality assurance (it's great)


lwm:
https://github.com/klen/pylama

This in combination with https://github.com/OpenIDC/pyoidc/issues/297 make for nice guarantees on style.

- [x] "*" imports and unused imports #305 
- [x] unused local variable #306
- [x] whitespace errors, blank lines errors #307
- [x] other errors and travis integration #308
- [ ] mccabe score below 15

lwm:
This will certainly require a heroic effort :beers:

lwm:
Closing in favour of https://github.com/OpenIDC/pyoidc/issues/311.

PR #305, opened 2017-03-28
Cleanup imports

tpazderka:
Part 1 of #298


lwm:
go go go go go go go go go go :+1:

tpazderka:
I have just realized that this might cause breakage on existing user code.
The previous availability of `oic.oauth2.message` from within `oic.oath2` might sause some issues.

@rohe I am not sure if this was by design or not but we should at least note in the changelog.

lwm:
> The previous availability of oic.oauth2.message from within oic.oath2 might sause some issues.

Oh! You are right. Please make a change log entry and bump the version as a major change, because this is breaking (maybe, in some cases ;)). I'll make a pypi release there after. Good catch.

PR #308, opened 2017-03-28
Pylama

tpazderka:
part 3 of #298

### DIFF
--- a/src/oic/oic/claims_provider.py
+++ b/src/oic/oic/claims_provider.py
@@ -96,7 +96,6 @@ class ClaimsServer(Provider):
         logger.info("RESPONSE: %s" % (sanitize(cresp.to_dict()),))
         return cresp
 
-    # noinspection PyUnusedLocal
     def _distributed(self, info):
         # store the user info so it can be accessed later
         access_token = rndstr()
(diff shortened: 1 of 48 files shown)


tpazderka:
Ok, I think I am done here

src/oic/oic/claims_provider.py
```diff
@@ -96,7 +96,7 @@ def _aggregation(self, info):
         logger.info("RESPONSE: %s" % (sanitize(cresp.to_dict()),))
         return cresp
 
-    #noinspection PyUnusedLocal
+    # noinspection PyUnusedLocal
```
lwm: I was wondering if we shouldn't just get rid of these 'no qa' comments and fix up the issues they relate to?

@rohe, you probably have something to add there?

src/oic/utils/userinfo/ldap_info.py
```diff
@@ -86,10 +86,12 @@ def __call__(self, userid, client_id, user_info_claims=None,
...
         try:
             res = self.ld.search_s(*arg)
-        except:
+        except Exception:
+            # FIXME: This shoudl catch specific exception from `self.ld.search_s()`
             try:
                 self.ld.close()
-            except:
+            except Exception:
+                # FIXME: This shoudl catch specific exception from `self.ld.close()`
```
lwm: shoudl -> should

src/oic/oic/claims_provider.py
rohe: We should.

tpazderka: OK then, I will go through and remove those.

tpazderka: Removed from `src/*` and `tests/*`. It turned out that those issues are already resolved.

lwm:
Great, just squash that fixup and you're good to go! :beer:

PR #312, opened 2017-03-30
Adding __eq__ method to KeyJar class

rohe:
Added a method that compares two KeyJar instances for equality.


rohe:
Why does Travis fail on:
ERROR: InvocationError: '/usr/bin/make check-isort'
??

tpazderka:
Because your imports are incorrectly sorted.

src/oic/utils/keyio.py
```diff
@@ -776,6 +776,36 @@ def get_issuer_keys(self, issuer):
...
+        # number of issuer_keys must be equal
+        if self.issuer_keys.keys() != other.issuer_keys.keys():
+            return False
+
+        for issuer in self.issuer_keys.keys():
+            mine = self.get_issuer_keys(issuer)
+            others = other.get_issuer_keys(issuer)
+            if len(mine) != len(others):
+                return False
+            for a_key in mine:
```
lwm: I think you could replace this nested for logic as `return set(mine) & set(others)`, no?

https://docs.python.org/3/library/stdtypes.html#set

lwm:
You'll need to clean up your commit log. This should be a single commit.

PR #317, opened 2017-03-31
In Python 2.7, this conversion to `newstr` triggers:

VJalili:
AttributeError: 'newstr' object has no attribute 'items'


lwm:
Thanks @VJalili! You're getting the following code quality failure:

```
quality runtests: commands[1] | make check-pylama
pylama src/ tests/
src/oic/oauth2/provider.py:7:1: W0611 'future.types.newstr' imported but unused [pyflakes]
make: *** [check-pylama] Error 1
```

I'm not really sure about this and would ask @rohe to take a look. Maybe you can just remove the code altogether if this is not something we need to support? Not sure.

tpazderka:
If the import is not used, then I would remove it. Also the commented code shouldn't be there.

What I am a little bit worried is that this is probably not a root cause of the issue. The `http.cookies` import should really come from `future.backports` library. That should take care of all of this.

VJalili:
This change has solved my problem, maybe it would help others run the examples. 
It's a patch, and might be useful till you find a better solution. 
Regarding the unused import: I'm focused on bigger issues running pyoidc, and forget to remove unused import. You can delete it if you want.

rohe:
Don't remember when newstr was added, was some time ago to fix a Python2 problem (obviously).
If that isn't the case anymore I'm happy with having it removed.

tpazderka:
I do believe that this fixes your issue, but we should make sure, that it doesn't break under python 3.

So proper tests should be added as well.

VJalili:
The if condition is met under python 2 (it reads `if PY2`). So, I believe this does not effect python 3.

tpazderka:
Ah, sorry, my bad. But the tests part still applies...

VJalili:
Updated the code to meet "checks" requirement.

VJalili:
@rohe if there is no further requirement on this PR, could you please merge it?

gene1wood:
I just encountered this bug in v0.10.0. and had a tough time figuring out that it was fixed in `master`. You may want to update [the changelog](https://github.com/OpenIDC/pyoidc/blob/master/CHANGELOG.md) to reflect this bug and this fix in the `v0.10.0.1 [UNRELEASED]` section

Issue #311, opened 2017-03-29
Decrease mccabe score


tpazderka:
Part 5 of #298

It might be a good idea to tackle them in small batches to decrease the possibility for bugs and oversights.

Behavior of public methods must remain stable.

-----
Top offenders:
```
src/oic/utils/authn/saml.py:25:1: C901 'TryExcept 25' is too complex (75)
src/oic/oic/provider.py:1319:1: C901 'Provider.do_client_registration' is too complex (39)
src/oic/utils/rp/__init__.py:99:1: C901 'Client.callback' is too complex (32)
src/oic/oauth2/message.py:378:1: C901 'Message._add_value' is too complex (31)
src/oic/oauth2/message.py:542:1: C901 'Message.from_jwt' is too complex (29)
src/oic/oauth2/provider.py:462:1: C901 'Provider.do_auth' is too complex (28)
src/oic/utils/keyio.py:472:1: C901 'KeyJar.get' is too complex (28)
src/oic/oauth2/message.py:157:1: C901 'Message.to_urlencoded' is too complex (27)
src/oic/oauth2/provider.py:345:1: C901 'Provider.auth_init' is too complex (27)
src/oic/oauth2/__init__.py:509:1: C901 'Client.parse_response' is too complex (27)
src/oic/oauth2/provider.py:666:1: C901 'Provider._complete_authz' is too complex (26)
```
Minor offenders
```
src/oic/extension/provider.py:449:1: C901 'Provider.client_info_endpoint' is too complex (19)
src/oic/extension/signed_http_req.py:127:1: C901 'SignedHttpRequest.verify' is too complex (18)
src/oic/extension/token.py:41:1: C901 'JWTToken.__call__' is too complex (24)
src/oic/oauth2/message.py:235:1: C901 'Message.from_urlencoded' is too complex (18)
src/oic/oauth2/message.py:492:1: C901 'Message.get_verify_keys' is too complex (21)
src/oic/oauth2/message.py:673:1: C901 'Message.verify' is too complex (17)
src/oic/oauth2/provider.py:214:1: C901 'Provider._verify_redirect_uri' is too complex (19)
src/oic/oauth2/util.py:88:1: C901 'set_cookie' is too complex (20)
src/oic/oauth2/util.py:168:1: C901 'verify_header' is too complex (16)
src/oic/oic/consumer.py:199:1: C901 'Consumer.begin' is too complex (19)
[...]

PR #318, opened 2017-03-31
Decrease complexity of authn/saml.py

tpazderka:

---
Ref #311


lwm:
Hey, I thought we were going to make these small!? A 300+ line diff is not easy to review ;)

tpazderka:
The only change is that the `try: except` moved to the top and everything got dedented...

tpazderka:
I do recommend doing the review from cmommandline using `git diff --ingore-space-change` :)

lwm:
Ah yes, thanks (didn't know about that option!)! Yes, I agree with this change, LGTM :+1:

PR #321, opened 2017-04-04
Moved sector_identifier checks outside

tpazderka:
Refactoring of `oic.oic.provider.Provider.do_clinet_registration` in order to decrease McCabe score.

---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -96,6 +96,10 @@ class InvalidRedirectURIError(Exception):
     pass
 
 
+class InvalidSectorIdentifier(Exception):
+    pass
+
+
 def devnull(txt):
     pass
 
@@ -1354,44 +1358,10 @@ class Provider(AProvider):
                                 status="400 Bad Request")
 
         if "sector_identifier_uri" in request:
-            si_url = request["sector_identifier_uri"]
             try:
-                res = self.server.http_request(si_url)
-            except ConnectionError as err:
-                logger.error("%s" % err)
-                return error_response(
-                    "invalid_configuration_parameter",
-                    descr="Couldn't open sector_identifier_uri")
-
-            if not res:
-                return error_response(
-                    "invalid_configuration_parameter",
-                    descr="Couldn't open sector_identifier_uri")
-
-            logger.debug("sector_identifier_uri => %s" % sanitize(res.text))
-
-            try:
-                si_redirects = json.loads(res.text)
-            except ValueError:
-                return error_response(
-                    "invalid_configuration_parameter",
-                    descr="Error deserializing sector_identifier_uri "
-                          "content")
-
-            if "redirect_uris" in request:
-                logger.debug("redirect_uris: %s" % request["redirect_uris"])
-                for uri in request["redirect_uris"]:
-                    try:
-                        assert uri in si_redirects
-                    except AssertionError:
-                        return error_response(
-                            "invalid_configuration_parameter",
-                            descr="redirect_uri missing from "
-                                  "sector_identifiers"
-                        )
-
-            _cinfo["si_redirects"] = si_redirects
-            _cinfo["sector_id"] = si_url
+                _cinfo["si_redirects"], _cinfo["sector_id"] = self._verify_sector_identifier(request)
+            except InvalidSectorIdentifier as err:
+                return error_response("invalid_configuration_parameter", descr=err)
         elif "redirect_uris" in request:
             if len(request["redirect_uris"]) > 1:
                 # check that the hostnames are the same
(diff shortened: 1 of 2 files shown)


src/oic/oic/provider.py
```diff
@@ -1354,44 +1357,11 @@ def do_client_registration(self, request, client_id, ignore=None):
...
-                            "invalid_configuration_parameter",
-                            descr="redirect_uri missing from "
-                                  "sector_identifiers"
-                        )
-
-            _cinfo["si_redirects"] = si_redirects
-            _cinfo["sector_id"] = si_url
+                _cinfo["si_redirects"], _cinfo["sector_id"] = self._verify_sector_identifier(request)
+            except InvalidSectorIdentifier as err:
+                return error_response("invalid_configuration_parameter",
```
lwm: Suppose we can make this all on one line? Could be nitpicking ;)

src/oic/oic/provider.py
```diff
@@ -1433,12 +1403,9 @@ def do_client_registration(self, request, client_id, ignore=None):
...
-            try:
-                n_keys = len(self.keyjar[client_id])
-                msg = "found {} keys for client_id={}"
-                logger.debug(msg.format(n_keys, client_id))
-            except KeyError:
-                pass
+            n_keys = len(self.keyjar.get(client_id, []))
+            if n_keys:
+                logger.debug("Found %s keys for client_id=%s", n_keys, client_id)
         except Exception as err:
```
lwm: Can we be more specific than `Exception` for the  `self.keyjar.load_keys(request, client_id)` failure when we catch?

src/oic/oic/provider.py
```diff
@@ -1501,6 +1468,39 @@ def verify_redirect_uris(registration_request):
...
+
+        :param request: Provider registration request
+        :return: si_redirects, sector_id
+        :raises: InvalidSectorIdentifier
+        """
+        si_url = request["sector_identifier_uri"]
+        try:
+            res = self.server.http_request(si_url)
+        except ConnectionError as err:
+            logger.error("{}", err)
```
lwm: Can just be `logger.error(err)`?

src/oic/oic/provider.py
```diff
@@ -1501,6 +1468,39 @@ def verify_redirect_uris(registration_request):
...
+            res = None
+
+        if not res:
```
lwm: This `raise` could be moved into the `except` block above?

src/oic/oic/provider.py
```diff
@@ -1501,6 +1468,39 @@ def verify_redirect_uris(registration_request):
...
+
+        logger.debug("sector_identifier_uri => %s", sanitize(res.text))
+
+        try:
+            si_redirects = json.loads(res.tezt)
+        except ValueError:
+            raise InvalidSectorIdentifier("Error deserializing sector_identifier_uri content")
+
+        if "redirect_uris" in request:
+            logger.debug("redirect_uris: %s". request["redirect_uris"])
```
lwm: What's that full stop doing!? Is this some new cool syntax that I didn't know about? :smile:

src/oic/oic/provider.py
```diff
@@ -1501,6 +1468,39 @@ def verify_redirect_uris(registration_request):
...
+            for uri in request["redirect_uris"]:
```
lwm: Might be just my preference but I like to skip nested fors (feel free to ignore):

``` python
if not all([uri in request["redirect_uris"] for uri in si_redirects]):
    raise InvalidSectorIdentifier("redirect_uri missing from sector_identifiers")
```

lwm:
> I would like to add some test to be sure it is working properly

:beers: :rainbow: :wine_glass: :rainbow_flag: :beers:

src/oic/oic/provider.py
tpazderka: Typos... typos everywhere...

src/oic/oic/provider.py
tpazderka: We should, I will look into this.

src/oic/oic/provider.py
tpazderka: Probably not, if you look at the previous code, it was doing the same thing. This catches the case when the response doesn't raise an exception but returns `None`.

src/oic/oic/provider.py
tpazderka: I always find dict comprehension unreadable (personal thing) so I will gladly go with the ignore on this one :)

src/oic/oic/provider.py
tpazderka: Mental note to myself: `KeyIOError` and `UnknownKeyType` are clearly visible as raised exceptions from called methods. Other could be possibly hidden.

tpazderka:
Tests for the method added. It is marked as internal since this is something that shouldn't be changed.

The score went down by `7` so some more work is needed...:)

tpazderka:
@lwm, could you please have a look so we can merge this?

Issue #322, opened 2017-04-04
what is `symkey` and where it is used?


VJalili:
The `Provider` class (@ /pyoidc/src/oic/oauth2/provider.py`) is initialized with a parameter called `symkey`.  This parameter is used to initialize a local parameter of the same name `self.symkey = symkey`.

But it seems `self.symkey` is never used. Am I missing something?

rohe:
It's used in a special type of user authentication.
SymKeyAuthn in src/oic/utils/user.py
We had a use case very early on.

lwm:
Hmmmm, `symkey` is not an optional parameter. I think we should make it optional.

rohe:
Agree

VJalili:
It is optional: `symkey="" `
I'm using `UserPassword` authentication, and I was wondering what is `SYM_KEY = "SoLittleTime,Got"` in the configuration. 
It's good to know that it's just used in `SymKeyAuthn`.

lwm:
> It is optional: symkey=""

Not for `oic.oic.Provider`. I'll submit a pull request.

PR #324, opened 2017-04-05
Make symkey an optional kwarg.

lwm:

---

Follows from https://github.com/OpenIDC/pyoidc/issues/322.

### DIFF
--- a/src/oic/oauth2/provider.py
+++ b/src/oic/oauth2/provider.py
@@ -152,7 +152,7 @@ class Provider(object):
     endp = [AuthorizationEndpoint, TokenEndpoint]
 
     def __init__(self, name, sdb, cdb, authn_broker, authz, client_authn,
-                 symkey="", urlmap=None, iv=0, default_scope="",
+                 symkey=None, urlmap=None, iv=0, default_scope="",
                  ca_bundle=None, verify_ssl=True, default_acr="",
                  baseurl='', server_cls=Server, client_cert=None):
         self.name = name
--- a/src/oic/oauth2/provider.py
+++ b/src/oic/oauth2/provider.py
@@ -10,8 +10,8 @@ import logging
 import os
 import sys
 import traceback
-
 from http.cookies import SimpleCookie
+
 from six import PY2
 
 from oic import rndstr
--- a/src/oic/oic/claims_provider.py
+++ b/src/oic/oic/claims_provider.py
@@ -69,7 +69,7 @@ class ClaimsServer(Provider):
     def __init__(self, name, sdb, cdb, userinfo, client_authn, urlmap=None,
                  ca_certs="", keyjar=None, hostname="", dist_claims_mode=None):
         Provider.__init__(self, name, sdb, cdb, None, userinfo, None,
-                          client_authn, "", urlmap, ca_certs, keyjar,
+                          client_authn, None, urlmap, ca_certs, keyjar,
                           hostname)
 
         if keyjar is None:
--- a/src/oic/utils/authn/user.py
+++ b/src/oic/utils/authn/user.py
@@ -11,6 +11,7 @@ import time
 
 import six
 
+from oic.exception import ImproperlyConfigured
 from oic.exception import PyoidcError
 from oic.oauth2 import compact
 from oic.utils import aes
@@ -387,6 +388,10 @@ class SymKeyAuthn(UserAuthnMethod):
 
     def __init__(self, srv, ttl, symkey):
         UserAuthnMethod.__init__(self, srv, ttl)
+
+        if symkey is not None and symkey == "":
+            msg = "SymKeyAuthn.symkey cannot be an empty value"
+            raise ImproperlyConfigured(msg)
         self.symkey = symkey
 
     def authenticated_as(self, cookie=None, authorization="", **kwargs):
--- a/src/oic/utils/http_util.py
+++ b/src/oic/utils/http_util.py
@@ -12,6 +12,7 @@ from six import PY2
 from six import text_type
 
 from oic import rndstr
+from oic.exception import ImproperlyConfigured
 from oic.exception import UnsupportedMethod
 from oic.utils import time_util
 from oic.utils.aes import decrypt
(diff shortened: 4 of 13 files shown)


VJalili:
Agree. I think the function signature is fine, but it's required to check for authentication method. Something link:
`if authenticationMethod == symkeyauthn  &&  symkey="" then 
  raise ImproperlyConfigured`

@lwm Could you please update your PR accordingly ?

tpazderka:
That probably is a side-effect of what I have pointed out. Empty string is used in creation of the cookie.

lwm:
Ah yes, I'll need to work a bit more on this and make some tests. Thanks for review :+1:

lwm:
Breaking changes! :bomb:

src/oic/oic/claims_provider.py
```diff
@@ -69,7 +69,7 @@ class ClaimsServer(Provider):
     def __init__(self, name, sdb, cdb, userinfo, client_authn, urlmap=None,
                  ca_certs="", keyjar=None, hostname="", dist_claims_mode=None):
         Provider.__init__(self, name, sdb, cdb, None, userinfo, None,
-                          client_authn, "", urlmap, ca_certs, keyjar,
+                          client_authn, rndstr(), urlmap, ca_certs, keyjar,
```
tpazderka: This can probably cause issues...

src/oic/oauth2/provider.py
```diff
@@ -153,9 +154,14 @@ class Provider(object):
     endp = [AuthorizationEndpoint, TokenEndpoint]
 
     def __init__(self, name, sdb, cdb, authn_broker, authz, client_authn,
-                 symkey="", urlmap=None, iv=0, default_scope="",
+                 symkey=None, urlmap=None, iv=0, default_scope="",
                  ca_bundle=None, verify_ssl=True, default_acr="",
                  baseurl='', server_cls=Server, client_cert=None):
+
+        if symkey is not None and symkey == "":
```
tpazderka: Maybe we can postpone this check till the point where the symkey is actually needed? Otherwise this check seems a little bit pointles since it passes for `None` which will break stuff.

tpazderka:
Sure, we can remmember to bump the version, but we should wait so we do not release that often.

tpazderka:
Also you have a lot of unrelated cleanup changes. I would rather see them in a separate issue targeted at cleanup.

lwm:
> Also you have a lot of unrelated cleanup changes. I would rather see them in a separate issue targeted at cleanup.

Yep, fair point, please see https://github.com/OpenIDC/pyoidc/pull/329.

src/oic/oic/claims_provider.py
lwm: Yes, you're right, will set to `None` instead.

src/oic/oauth2/provider.py
lwm: I see what you mean, I'll try to find the point where it is used.

lwm:
TODO:
  - [ ] Validate `symkey` is not the empty string in `CookieDealer` and `SymkeyAuthn`.
  - [ ] Write tests to cover these changes
  - [ ] Add some documentation (better start now than never)

lwm:
OK gentlemen, please review this again! I've given it another go :+1:

src/oic/utils/http_util.py
```diff
@@ -437,12 +438,18 @@ def __init__(self, srv, ttl=5):
...
 
     def init_srv(self, srv):
-        if srv:
-            self.srv = srv
-
-            for param in ["seed", "iv"]:
-                if not getattr(srv, param, None):
-                    setattr(srv, param, rndstr().encode("utf-8"))
+        if not srv:
+            return
```
lwm: Because the original code did not either and I wasn't sure if this was acceptable. I'll patch that up. Please, @rohe, can you confirm this?

src/oic/utils/authn/user.py
```diff
@@ -387,6 +388,10 @@ class SymKeyAuthn(UserAuthnMethod):
 
     def __init__(self, srv, ttl, symkey):
         UserAuthnMethod.__init__(self, srv, ttl)
+
+        if symkey is not None and symkey == "":
```
lwm: > The actual check is even stricter, due to the minimal aes keylength, that is enforced in oic.util.aes.build_cipher.
> Maybe the aes code could provide a method to check

I don't see the connection from the `Provider.symkey` to the `oic.util.aes.build_cipher`? Does one of the provider classes call this function? We put the check here [due to the idea in this comment](https://github.com/OpenIDC/pyoidc/pull/324#pullrequestreview-31133941).

lwm: > The actual check is even stricter, due to the minimal aes keylength, that is enforced in oic.util.aes.build_cipher. ... Maybe the aes code could provide a method to check

I don't see the connection from the `Provider.symkey` to the `oic.util.aes.build_cipher`? Does one of the provider classes call this function? We put the check here [due to the idea in this comment](https://github.com/OpenIDC/pyoidc/pull/324#pullrequestreview-31133941).

src/oic/utils/http_util.py
lwm: I'd actually be happier with making this a separate ticket if you don't mind @schlenk. I have preserved the original behaviour (I think!) in this change but your recommendation (which I think is totally valid!) will bring something new. I've created https://github.com/OpenIDC/pyoidc/issues/365 for that.

src/oic/utils/authn/user.py
schlenk: You are right (in a way). 

The only usecase for using the Provider.symkey currently is to use it as a key for AES encryption. And AES encryption needs a minimum key length (as checked in build_cipher). So it makes no sense to have a symkey that is too short to use as an AES key. (e.g. just 2 Bytes or something like that).

Currently, you can set a symkey like 'ab' and all the setup/init works, but you blow up at runtime, when you get FailedAuthentication or other AssertionErrors from aes.encrypt(). So it would be user friendlier to prevent a too short to be useful symkey too, instead of just empty and `None`.

src/oic/utils/http_util.py
schlenk: ok, fine with me.

src/oic/utils/authn/user.py
lwm: Aha! I see, I see. Thanks for the explanation. I'll follow up with a key length check.

rohe:
Over the time I've been working on this library I've gone back and forth between throwing an exception or returning something to signal that an error occurred. 
Lately I'm leaning more toward the 'throw an exception' camp so I would be OK with changing the code to raise an exception.

PR #325, opened 2017-04-05
Refactor claims_match.

lwm:

---

~~@rohe, you'll have to advise on this one. I was just passing through
here and started to try to figure out what this function is really
doing.~~

OK, I think I've covered this refactoring.

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -234,14 +234,12 @@ PREFERENCE2PROVIDER = {
     "id_token_encrypted_response_enc":
         "id_token_encryption_enc_values_supported",
     "default_acr_values": "acr_values_supported",
-    # "require_auth_time":"",
     "subject_type": "subject_types_supported",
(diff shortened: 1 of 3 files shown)


src/oic/oic/__init__.py
```diff
@@ -258,19 +258,17 @@ def add_token(self, resp):
...
-            if value not in val:
-                return False
-                # Whether it's essential or not doesn't change anything here
-    return True
+    :param claim1: First claim value
+    :param claim2: Second claim value
+    :return: Boolean
+    """
+    if isinstance(claim2, Iterable):
+        return claim1 in claim2
```
tpazderka: I think you need to do `claim1 in claim2.values()` to keep the logic of previous implementation.

lwm: Oh yes, better cover that too!

tpazderka:
A test is missing for the `Iterable` claim and for simple `str` instances. Otherwise it looks good.

lwm:
Missing tests added!

PR #338, opened 2017-05-02
Encodes unicode strings to ascii resolving an issue of using unicode keys in wrapper dictionary

VJalili:
- [x] no changes relevant to user.
- [x] does not effect documentation.

PR #343, opened 2017-05-04
Follow on from #342/#338

lwm:
Follow on from https://github.com/OpenIDC/pyoidc/pull/342 and https://github.com/OpenIDC/pyoidc/pull/338. @VJalili, can you weigh in here if this the set of changes you had originally? I would remove the comments? Also, should we not use `six` instead of all these `sys.version_info` checks?

### DIFF
--- a/oidc_example/op3/server.py
+++ b/oidc_example/op3/server.py
@@ -3,13 +3,16 @@ __author__ = 'Vahid Jalili'
 
 from future.backports.urllib.parse import parse_qs
 
+import argparse
+import importlib
 import json
 import os
 import re
 import sys
 import traceback
-import argparse
-import importlib
+
+from cherrypy import wsgiserver
+from cherrypy.wsgiserver.ssl_builtin import BuiltinSSLAdapter
 from mako.lookup import TemplateLookup
 
 from oic.oic.provider import AuthorizationEndpoint
@@ -27,17 +30,10 @@ from oic.utils.authn.user import UsernamePasswordMako
 from oic.utils.authz import AuthzHandling
 from oic.utils.http_util import *
 from oic.utils.keyio import keyjar_init
+from oic.utils.sdb import SessionDB
 from oic.utils.userinfo import UserInfo
 from oic.utils.webfinger import OIC_ISSUER
 from oic.utils.webfinger import WebFinger
-
-
-from cherrypy import wsgiserver
-from cherrypy.wsgiserver.ssl_builtin import BuiltinSSLAdapter
-
-from oic.utils.sdb import SessionDB
-
-
 
 LOGGER = logging.getLogger("")
 LOGFILE_NAME = 'oc.log'
--- a/oidc_example/op3/server.py
+++ b/oidc_example/op3/server.py
@@ -92,7 +92,6 @@ def key_rollover(self, environ, start_response, _):
     # expects a post containing the necessary information
     _txt = get_post(environ)
     _jwks = json.loads(_txt)
-    # logger.info("Key rollover to")
     provider.do_key_rollover(_jwks, "key_%d_%%d" % int(time.time()))
     # Dump to file
     f = open(jwksFileName, "w")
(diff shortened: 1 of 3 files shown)


VJalili:
@lwm Thanks for the cleanup. 

> can you weigh in here if this the set of changes you had originally?

Correct. Lets merge these changes. 


> Also, should we not use six instead of all these sys.version_info checks?

Thats a great alternative. If you prefer that, you may updated the PR.

oidc_example/op3/server.py
```diff
@@ -131,11 +131,13 @@ def __init__(self, provider, urls):
             (r'^.well-known/simple-web-discovery', self.swd_info),
             (r'^.well-known/host-meta.json', self.meta_info),
             (r'^.well-known/webfinger', self.webfinger),
+            #    (r'^.well-known/webfinger', webfinger),
```
tpazderka: The commented lines shouldn't be here.

src/oic/utils/shelve_wrapper.py
```diff
@@ -16,31 +18,46 @@ def __len__(self):
         return db.__len__()
 
     def has_key(self, key):
+        if sys.version_info < (3, 0) and not isinstance(key, str):
```
tpazderka: These branches are never called in tests. Also the `six` should be used as suggested in comments.

lwm:
OK, finally came back to this one. Didn't have time to write tests (could merge now and add issue to remember - @VJalili, feel free since these were your original changes).

src/oic/utils/shelve_wrapper.py
```diff
@@ -16,31 +19,46 @@ def __len__(self):
...
+        if six.PY2 and not isinstance(key, six.string_types):
+            key = self._normalizeString(key)
         db = self._reopen_database()
         db.__delitem__(key)
 
     def _reopen_database(self):
         return shelve.open(self.filename, writeback=True)
 
+    def _normalizeString(self, key):
+        return unicodedata.normalize('NFKD', key).encode('ascii', 'ignore')
```
schlenk: This looks wrong for two reasons:

1) Open ID Connect Core explicitly forbids Unicode Normalization for string comparision
https://openid.net/specs/openid-connect-core-1_0.html#StringOps

2) Is encode('ascii') correct here? I would expect utf-8.

A malicious attacker could send unicode access tokens that degenerate to some totally different 7-bit ASCII string when treated with this. Not sure if that would be an issue though, but in theory this could lead to client_id confusion.

For that reason i would suggest to just try an encode 'utf-8' without normalization and just throw an exception if it fails.

VJalili:
I think I did the unicode normalization because of the issues I was facing running in python 2.7. @schlenk if such conversions are "forbidden", then we might need to avoid running pyoidc under python 2.7. This is mainly because of differences between string representation in python 2.7 and 3.x. 

However, being only python 3.x compatible, is a bit strong requirement as it could exclude a considerable portion of the users. If we could not use this approach, then I would suggest thinking of an alternative (and recommended) backward compatible implementation.

lwm:
@VJalili, I'm going to close off this PR now. I'm not really sure about the changes, since I was just supposed to clean them up in the end, and it seems there are new changes to be made. If you're up for it, please re-submit in a new PR. Thanks.

src/oic/utils/shelve_wrapper.py
lwm: Thanks for review! I defer to @VJalili on this one (see https://github.com/OpenIDC/pyoidc/pull/343#issuecomment-305991554 below).

VJalili:
You can close this, however, I don't know how you want to solve the issue otherwise.

schlenk:
@VJalili Just doing an `unicode(key).encode('utf-8')` is fine, as it is losless and bidirectional. And works for Python 2.7 just fine (using exactly that myself). But normalization looses information so two different things can map to the same normalized string. 

This can be a security problem if normalization and comparision is not done in the same way on all code paths or system, so OIDC spec simply states it should not be done to avoid those problems.

VJalili:
@schlenk  interesting! would you mind update the code accordingly?

lwm:
> I don't know how you want to solve the issue otherwise.

You started this change set, so you can do it :smile:

PR #347, opened 2017-05-04
Issuer with path

### LINKED ISSUE #346: Using an issuer with a path component produces invalid openid-configuration URLs
Assume an issuer retrieved via Client.discovery() contains a path component (which seems legal from my reading of section 3 of the Open ID Connect Discovery 1.0 spec, which forbids fragments and query parts but not path elements in the issuer):
e.g. 

    issuer = 'https://example.com/oidc/'
    pinfo = client.provider_config(issuer)

=> Error, the URL requested is 'https://example.com/oidc/.well-known/openid-configuration'

I think it should be https://example.com/.well-known/openid-configuration at least according to RFC 5785 all the `/.well-known` URLs start at the top level.

The problem is the naive url construction in `provider_config`, which should probably be done like this instead:

    _authority = "{0.scheme}://{0.netloc}".format(urlparse(_issuer))
    url = serv_pattern % _authority

schlenk:
Allow an issuer with a path in the provider_config method.

This normalizes the issuer down to scheme + netloc and fixes #346.

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -970,12 +970,13 @@ class Client(oauth2.Client):
     def provider_config(self, issuer, keys=True, endpoints=True,
                         response_cls=ProviderConfigurationResponse,
                         serv_pattern=OIDCONF_PATTERN):
-        if issuer.endswith("/"):
-            _issuer = issuer[:-1]
-        else:
-            _issuer = issuer
 
-        url = serv_pattern % _issuer
+        # Issuer MAY contain a path element, but webfinger .well-known URLs
+        # are always rooted at the top (RFC 5785), so this must be normalized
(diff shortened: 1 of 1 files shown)


lwm:
Thanks @schlenk and welcome! Can you please provide a test to go along with this change?

tpazderka:
Just to be sure, could @rohe look at this as well?

If I understand this correctly, this is a move in correct direction, but I would like to be sure :)

tpazderka:
Also it should be noted in Changelog since this is changing the way the discovery works.

rohe:
I'm boarding a plane in a few minutes on my way home from the US.
Will have a look at this during the weekend.

schlenk:
The assumptions here are wrong, see the linked issue.

PR #351, opened 2017-05-18
Status code can be 400, 401, 403 or .. !

rohe:
According to RFC6749 an OP shall return a 400 when it encounters an error in processing a request.
The OIDC core document in section 5.3.3 in the example uses 401.
During conformance testing I've also seen 403's in some cases.
So, there is reason to widen then scope for possible status_code denoting protocol errors.

The fact that I've seen software return a 200 with an error message is another matter and is already covered in the code.

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -849,7 +849,7 @@ class Client(oauth2.Client):
                 sformat = "jwt"
         elif resp.status_code == 500:
             raise PyoidcError("ERROR: Something went wrong: %s" % resp.text)
-        elif resp.status_code == 400:
+        elif 400 <= resp.status_code < 500 :
             # the response text might be a OIDC message
             try:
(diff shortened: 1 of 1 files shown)


src/oic/oic/__init__.py
```diff
@@ -849,7 +849,7 @@ def do_user_info_request(self, method="POST", state="", scope="openid",
                 sformat = "jwt"
         elif resp.status_code == 500:
             raise PyoidcError("ERROR: Something went wrong: %s" % resp.text)
-        elif resp.status_code == 400:
+        elif 400 <= resp.status_code < 500 :
```
tpazderka: There is a space before the `:`.

Otherwise I am fine with that.

rohe:
Removed the space.
Also fixed a reference to a exception class that was wrong.
Actually the class in question is defined in 2 places.
Should fix that but that's the topic for another PR.

Issue #349, opened 2017-05-08
oic.utils.sdb.Crypt uses a weak key derivation function and constant IV


schlenk:
The oid.utils.sdb.Crypt function code uses a single round of sha256 as a rather weak KDF.

Additionally it uses a constant Initialization Vector for AES.CBC which is also not a great idea.
(see for example http://cwe.mitre.org/data/definitions/329.html  and for a more visual explanation of the attack, see http://stackoverflow.com/questions/3008139/why-is-using-a-non-random-iv-with-cbc-mode-a-vulnerability ).

Well even using CBC mode is not a good idea anymore. Should be some form of AEAD (like AES-GCM, or similar).

tpazderka:
Thank you for your suggestion. Could you please submit a PR?

Victor-Savu:
Is this issue resolved now? Can it be closed?

schlenk:
This is mostly fixed, yes. The only part remaining is the weak KDF, but thats actually more a bad naming for the 'password' parameter, which should not be a password (needing a KDF to add entropy) but should simply be a high entropy key already.
So yes, okay to close.

lwm:
> but thats actually more a bad naming for the 'password' parameter, which should not be a password (needing a KDF to add entropy) but should simply be a high entropy key already.

If you feel like this should change, please open an issue. All breaking changes considered ;)

> So yes, okay to close.

Closing then :beers:

PR #354, opened 2017-05-19
Better token crypto

schlenk:
This fixes the constant IV issue in #349 by using the fernet scheme to do the encryption.

This does not yet add a KDF function to stretch a weak password into some stronger key. This is okay if the 'password' is not really a password but a high entropy key. 

It is still 'padding compatible' with the old Crypt code, even if the benefits of manual padding to the AES blocksize seems doubtful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,14 @@ The format is based on the [KeepAChangeLog] project.
 - [#318]: `oic.utils.authn.saml` raises `ImportError` on import if optional `saml2` dependency is not present.
 - [#325]: `oic.oic.claims_match` implementation refactored.
 
+### Security
+- [#349]: Changed crypto algorithm used by `oic.utils.sdb.Crypt` for token encryption to Fernet. Old stored tokens are incompatible. 
+
 [#313]: https://github.com/OpenIDC/pyoidc/issues/313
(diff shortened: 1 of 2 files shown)


rohe:
I'm OK with this. Anyone against ?

tpazderka:
I am OK with this as well.

@schlenk Could you please make it a single commit and add a note in Changelog since it renders old tokens unusable (if I am not mistaken).

schlenk:
@tpazderka You are right, the update changes the encrypted format of the token. So old stored tokens would need a migration (decrypt with old code, encrypt with new) or need to be reissued.

Added the changelog entry under Security, as suggested by keepachangelog.

CHANGELOG.md
```diff
@@ -15,10 +15,14 @@ The format is based on the [KeepAChangeLog] project.
 - [#318]: `oic.utils.authn.saml` raises `ImportError` on import if optional `saml2` dependency is not present.
 - [#325]: `oic.oic.claims_match` implementation refactored.
 
+### Security
+- [#349]: Changed crypto algorithm used by `oic.utils.sdb.Crypt` for token encryption to Fernet. Old stored tokens are incompatibel.
```
tpazderka: Typo `incompatible`.

PR #358, opened 2017-05-31
Fixed claims_match and added tests.

rohe:
Modified claims_match to correctly implement what's specified in section 5.5.1 of http://openid.net/specs/openid-connect-core-1_0.html.
Also added tests to increase coverage and make it clearer what the function is expected to do.

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -258,21 +258,41 @@ PARAMMAP = {
 }
 
 
-def claims_match(claim1, claim2):
+def claims_match(value, claimspec):
     """
-    Is claim1 equal to or present in claim2?
+    Implements matching according to section 5.5.1 of 
+    http://openid.net/specs/openid-connect-core-1_0.html 
+    The lack of value is not checked here.
+    Also the text doesn't prohibit having both 'value' and 'values'.
 
-    :param claim1: First claim value
-    :param claim2: Second claim value
+    :param value: single value or list of values 
+    :param claimspec: None or dictionary with 'essential', 'value' or 'values'
+    as key
     :return: Boolean
     """
-    if claim2 is None:
+    if claimspec is None:  # match anything
         return True
-    if isinstance(claim2, dict):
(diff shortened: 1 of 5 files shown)


rohe:
What ? How can adding an empty line cause this massive Travis CI failure ?

tpazderka:
`six` is no longer being installed in Travis...

Don't know why but we need to fix it on `master` as well.

tpazderka:
OK, it looks like a fault on Travis side... I have restarted the failed builds and they are passing now...

tests/test_claims.py
```diff
@@ -24,7 +24,17 @@ def test_non_aggregate_claims():
...
+                "values": ["urn:mace:incommon:iap:silver",
+                           "urn:mace:incommon:iap:bronze"]}
+    }
+
+    assert claims_match("248289761001", claims_request['sub'])
+    assert claims_match("123456789012", claims_request['sub']) is False
+    assert claims_match("123456789", claims_request['auth_time'])
+    assert claims_match("urn:mace:incommon:iap:silver", claims_request['acr'])
+    assert claims_match("urn:mace:incommon:iap:gold",
+                        claims_request['acr']) is False
```
tpazderka: I am missing a test for a list of values.

src/oic/oic/__init__.py
```diff
@@ -258,21 +258,41 @@ def add_token(self, resp):
...
+
+    matched = False
+    for key, val in claimspec.items():
+        if key == "value":
+            if value == val:
+                matched = True
+        elif key == "values":
+            if value in val:
+                matched = True
+        elif key == 'essential':
```
tpazderka: This branch of the condition can be probably omitted.

rohe: No, if we omit it then the {'essential': True/False} case will fail.
{'essential': True/False} in essence is the same as value == *. 
That is, as long as there is a value it matched.
This means that 'essential' together with 'value'/'values' doesn't add anything. It's only if it's by itself that it matters.

Issue #363, opened 2017-06-05
CookieDealer.create_cookie reuses the IV for AES encryption of cookies


schlenk:
The CookieDealer class has some peculiar use of the AES IV.

It creates ONE IV when the class is initialized, this is then reused for every cookie created by the class.
Depending on AES mode, this can be fatal for the security of the key. For AES-GCM it would totally ruin security, for CBC mode it is not as fatal, but still bad.

The Cookie dealer should probably use one fresh IV per cookie and store it in the cookie with the encrypted value.

lwm:
Ah, I see! Well spotted!

PR #366, opened 2017-06-05
Assure fresh IV for CookieDealer

lwm:
- [ ] ~~The documentation has been updated, if necessary.~~

---

Closes https://github.com/OpenIDC/pyoidc/issues/363.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on the [KeepAChangeLog] project.
 
 ## 0.10.0.1 [UNRELEASED]
 ### Fixed
+- [#366]: Assure fresh seed/iv for each cookie from CookieDealer
 - [#358]: Fixed claims_match
 - [#313]: Catch exception correctly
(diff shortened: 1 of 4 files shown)


lwm:
Oh! If the CookieDealer creates a cookie, it can parse it afterwards (because it stores the seed and IV on the class). However, if the CookieDealer creates two cookies, it can only parse the second one and not the first (because this change generates new seed/IV values).

What's to be done about that?

schlenk:
@lvm The same as always with IV's. Those are called IVs and not keys for a reason, IVs do not need to be kept not secret, so you can simply prepend them to the message. So just prefix the cookie contents with the IV. (the SymKeyAuthn does that already).

lwm:
Oof, actually, making this change is pretty serious (mega test failure). I'll leave this for now.

schlenk:
Ok, I'll have a look at it hopefully soon. There are some other things that should be done better with cookie encryption (e.g. AEAD encryption instead of simple AES-CBC to prevent attackers messing with message integrity).

PR #371, opened 2017-06-07
Use find_packages

lwm:
Let's try this again :)


lwm:
Hmm, I have no idea why that extra commit is there!? Also, this isn't working :smile:

PR #373, opened 2017-06-07
Improve cookie crypto for CookieDealer

### LINKED ISSUE #363: CookieDealer.create_cookie reuses the IV for AES encryption of cookies

schlenk:
Added per cookie IV's to the CookieDealers encryption handling.

This fixes #363.

Also restyled the encrypt and MAC construction for cookie security to
use a more modern AEAD approach.

In this case it is AES-SIV (RFC  5297), which has the nice property to
be a bit resistant to IV reuse.

### DIFF
--- a/src/oic/utils/aes.py
+++ b/src/oic/utils/aes.py
@@ -111,7 +111,7 @@ class AEAD(object):
     """
     Authenticated Encryption with Associated Data Wrapper
 
-    This does encrypts and does an integrity check in one
+    This does encryption and integrity check in one
     operation, so you do not need to combine HMAC + encryption
     yourself.
 
--- a/src/oic/utils/aes.py
+++ b/src/oic/utils/aes.py
@@ -7,7 +7,9 @@ from base64 import b64encode
 
 from Cryptodome import Random
 from Cryptodome.Cipher import AES
+from six import binary_type
 from six import indexbytes
+from six import text_type
 
 __author__ = 'rolandh'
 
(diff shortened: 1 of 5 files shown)


rohe:
LGTM apart from the import sorting :-/

schlenk:
Seems isort messed up the line endings..., need to fix that again.

src/oic/utils/aes.py
```diff
@@ -105,6 +107,87 @@ def decrypt(key, msg, iv=None, padding="PKCS#7", b64dec=True):
     return res.decode("utf-8")
 
 
+class AEAD(object):
+    """
+    Authenticated Encryption with Associated Data Wrapper
+
+    This does encrypts and does an integrity check in one
```
lwm: `%s/does encrypts and does an/does encryption and an/g`

src/oic/utils/aes.py
```diff
@@ -105,6 +107,87 @@ def decrypt(key, msg, iv=None, padding="PKCS#7", b64dec=True):
...
+
+    For the default SIV mode, you need one of:
+
+        - 256-bit key, 128-bit IV to use AES-128
+        - 384-bit key, 192-bit IV to use AES-192
+        - 512-bit key, 256-bit IV to use AES-256
+
+    """
+    def __init__(self, key, iv, mode=AES.MODE_SIV):
+        assert isinstance(key, binary_type)
```
lwm: Should we not raise an `ImproperlyConfigured` error here for consistency?

src/oic/utils/http_util.py
```diff
@@ -262,34 +265,110 @@ def _expiration(timeout, time_format=None):
...
     if not timestamp:
         timestamp = str(int(time.time()))
-    signature = cookie_signature(seed, load.encode("utf-8"),
-                                 timestamp.encode("utf-8"))
-    cookie[name] = "|".join([load, timestamp, signature])
+
+    bytes_load = load.encode("utf-8")
+    bytes_timestamp = timestamp.encode("utf-8")
+
+    # If we have an encryption key, we use an AEAD cipher instead of
```
lwm: I would be worried that the comments will not match the implementation if changes come. Can you make a more higher level description of the function in the docstring?

tests/test_aes.py
```diff
@@ -16,3 +20,41 @@ def test_encrypt_decrypt():
     encrypted_msg = encrypt(key_, msg_, 0)
     txt = decrypt(key_, encrypted_msg, 0)
     assert txt == msg_
+
+
+def test_AEAD_good():
+    key = os.urandom(32)
```
lwm: Do you want to take advantage of our new and shiny `conftest.py` and add what you can (seed, iv, cleartext) as fixtures?

src/oic/utils/http_util.py
schlenk: I'll try.

src/oic/utils/aes.py
schlenk: This assert is mostly there to help with PY2/PY3 issues so keys/ivs etc stay 'byte strings'.
The configuration mistakes possible should be handled higher up, down here it is a programming error.

tests/test_aes.py
schlenk: Will have a look if it fits.

src/oic/utils/aes.py
lwm: Got it :+1:

tests/test_aes.py
schlenk: I don't think its useful in the main conftest.py file, but i moved them to local fixtures in the test file.

schlenk:
@lwm I fixed the docstrings a bit, so maybe its good to go now.

tests/test_aes.py
lwm: Sure, fine for now. If we need them, they can move.

lwm:
> @lwm I fixed the docstrings a bit, so maybe its good to go now.

:+1:

PR #374, opened 2017-06-08
to_jwe/from_jwe fix

rohe:

to_jwe/from_jwe using elleptic curve crypto is dependent on pyjwkest to be of version 1.3.3 or later to work.

### DIFF
--- a/tests/test_oauth2_message.py
+++ b/tests/test_oauth2_message.py
@@ -30,8 +30,18 @@ from oic.oauth2.message import TooManyValues
 from oic.oauth2.message import json_deserializer
 from oic.oauth2.message import json_serializer
 from oic.oauth2.message import sp_sep_list_deserializer
+from oic.utils.keyio import build_keyjar
 
 __author__ = 'rohe0002'
+
+keys = [
+    {"type": "RSA", "use": ["sig"]},
+    {"type": "RSA", "use": ["enc"]},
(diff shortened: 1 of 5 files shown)


rohe:
Actually pyjwkest 1.3.4 is better :-)

rohe:
Not really sure why Tarvis fails. Works on my machine.

lwm:
If you're relying on the new pyjwktest, it's being installed from the [requirements file](https://github.com/OpenIDC/pyoidc/blob/master/requirements/base.txt#L14). Try running a `make upgrade` and adding it to the PR.

We should probably have a discussion on how we want to support requirement.txt vs. setup.py installs. But for another day :)

rohe:
Damn, OK got it.

rohe:
OK, so the problem was with pyjwkest. Fixed that now.

tests/test_oauth2_message.py
```diff
@@ -595,3 +605,33 @@ def test_missing_required(self):
 
         with pytest.raises(MissingRequiredAttribute):
             err.to_urlencoded()
+
+
+def test_to_jwt_rsa():
+    msg = Message(a='foo', b='bar', c='tjoho')
```
lwm: This can be a fixture that goes in `conftest.py`.

lwm:
LGTM.

PR #375, opened 2017-06-08
base URL according to urljoin can only contain one path component.

rohe:
I have use cases where that is not true.


schlenk:
LGTM. 

And yes, guess there are some other instances where urljoin() isn't the right tool for the job. Probably always when baseurl contains a path.

lwm:
> I have use cases where that is not true.

Test case would be preferred :smile:

PR #377, opened 2017-06-09
Make sure the issuer of the ID Token is the correct one.

rohe:
Should not accept an ID Token from anyone but the expected one.

### DIFF
--- a/src/oic/oic/message.py
+++ b/src/oic/oic/message.py
@@ -870,7 +870,7 @@ class ProviderConfigurationResponse(Message):
         assert not parts.query and not parts.fragment
 
         if any("code" in rt for rt in self[
-            "response_types_supported"]) and "token_endpoint" not in self:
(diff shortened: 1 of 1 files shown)


rohe:
Travis fails because #374 isn't accepted yet.

src/oic/oic/message.py
```diff
@@ -674,6 +675,13 @@ class IdToken(OpenIDSchema):
     def verify(self, **kwargs):
         super(IdToken, self).verify(**kwargs)
 
+        try:
+            kwargs['iss'] == self['iss']
```
schlenk: Why should this produce an AssertionError? no assert anywhere...

It is also not a great idea to use 'assert' for this kind of code, as 'assert' gets removed when someone runs python with optimization (-O), so logic fails.

rohe: Correct, I should have know better.

lwm:
:beer:

PR #379, opened 2017-06-16
Kmk

rohe:
Changed the behaviour when a signed JWT is received that contains no kid while the JWKS of the issuer contains more then one key that could have been used.

The default behaviour is that the verification should fail when this happens.
This is what the standard specifies as the expected behaviour.

But if one wants it's now possible to specify issuers that are 'allowed' to not include a kid.
For those issuers the possible set of keys will be checked one at the time.

### DIFF
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -473,28 +473,51 @@ class Message(MutableMapping):
         _jws = JWS(self.to_json(lev), alg=algorithm)
         return _jws.sign_compact(key)
 
-    def _add_key(self, keyjar, issuer, key, key_type=''):
+    def _add_key(self, keyjar, issuer, key, key_type='', kid='',
+                 no_kid_issuer=None):
+
         try:
             logger.debug('Key set summary for {}: {}'.format(
                 issuer, key_summary(keyjar, issuer)))
         except KeyError:
             logger.error('Issuer "{}" not in keyjar'.format(issuer))
 
-        try:
-            kl = keyjar.get_verify_key(owner=issuer, key_type=key_type)
-        except KeyError:
-            pass
+        if kid:
+            _key = keyjar.get_key_by_kid(kid, issuer)
+            if _key and _key not in key:
+                key.append(_key)
+                return
         else:
-            for k in kl:
-                if k not in key:
-                    key.append(k)
+            try:
+                kl = keyjar.get_verify_key(owner=issuer, key_type=key_type)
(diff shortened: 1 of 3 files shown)


tests/test_oauth2_message.py
```diff
@@ -30,9 +30,35 @@
 from oic.oauth2.message import json_deserializer
 from oic.oauth2.message import json_serializer
 from oic.oauth2.message import sp_sep_list_deserializer
+from oic.utils.keyio import build_keyjar
 
 __author__ = 'rohe0002'
 
+keys = [
```
schlenk: All this setup should probably be in one or more pytest.fixture functions.

It clearly is a test fixture.

tests/test_oauth2_message.py
```diff
@@ -595,3 +621,100 @@ def test_missing_required(self):
 
         with pytest.raises(MissingRequiredAttribute):
             err.to_urlencoded()
+
+
+def test_to_jwt_rsa():
```
schlenk: Those tests could maybe be simplified into just one function with https://docs.pytest.org/en/latest/parametrize.html#

The body is nearly the same, just a few different parameters.

src/oic/oauth2/message.py
```diff
@@ -473,28 +473,55 @@ def to_jwt(self, key=None, algorithm="", lev=0):
...
+                        if allowed_kids:
+                            for k in kl:
+                                if k.kid in allowed_kids:
+                                    key.append(k)
+                        else:
+                            for k in kl:
+                                key.append(k)
 
     def get_verify_keys(self, keyjar, key, jso, header, jwt, **kwargs):
         try:
```
schlenk: docstring would be nice.

src/oic/oauth2/message.py
```diff
@@ -473,28 +473,55 @@ def to_jwt(self, key=None, algorithm="", lev=0):
...
+                    if kl[0] not in key:
+                        key.append(kl[0])
+                elif no_kid_issuer:
+                    try:
+                        allowed_kids = no_kid_issuer[issuer]
+                    except KeyError:
+                        return
+                    else:
+                        if allowed_kids:
+                            for k in kl:
```
lwm: :+1 and maybe change k->key, k1->verified for future us.

lwm:
Solid work. Please leave a change log entry too, should we consider a release after this?

rohe:
Yes we should do a new release !

Issue #326, opened 2017-04-06
Remove defaults for security related settings


tpazderka:
Some classes have security related settings as optional arguments with defined value or do define such value if `None` is provided. This is considered as a bad practice and either they should not be optional where it makes sense or should raise `ImproperlyConfiugured` if user tries to use them without providing a value.

lwm:
Sounds sensible!

schlenk:
At least:

- [ ] utils/sdb.py SessionDB `password` and `secret`

I don't think examples should matter in this case, if the secrets are in the configuration part.

Issue #146, opened 2015-11-23
oic.utils.SessionDB is not conducive to a shared DB with multiple SessionDB instances


gobengo:
It is common in my world to deploy multiple processes of an application server (like an oidc server) all sharing a database cluster. In this way, if one process or host crashes, the other instances can continue to serve requests.

I use `oic.utils.sdb` with a `db` object that reads/writes to a Redis instance on the network. I then run 2+ servers, each with a SessionDB instance that can read/write from that Redis.

However, I noticed that `SessionDB().gets_sids_for_sub()` does not return the same values on all hosts. It appears this is because this and a few other SessionDB methods make use of an in-memory `.uid2sid` cache that is only updated by writes to that one SessionDB instance. Writes to the underlying `.db` don't update that cache.

Here is a test suite that fails, illustrating that:

``` python
import unittest

from oic.oic.message import AuthorizationRequest
from oic.utils.sdb import AuthnEvent

from oic.utils.sdb import SessionDB


def create_auth_request():
    areq = AuthorizationRequest(response_type="code", client_id="client1",
                                redirect_uri="http://example.com/authz",
                                scope=["openid"], state="state000")
    return areq


def setup_session(session_db, uid):
    aevent = AuthnEvent(uid, 'salt')
    areq = create_auth_request()
    sid = session_db.create_authz_session(aevent=aevent,
                                          areq=areq)
    session_db.do_sub(sid, 'client_salt')
    return sid


class TestSessionDB(unittest.TestCase):

    def setUp(self):
        self.db = dict()
        self.sdb = SessionDB('https://1.com', db=self.db)

    def test_exists(self):
        self.assertTrue(bool(self.sdb))

    def test_get_sids_from_uid(self):
        # This passes
        uid = 'uid1'
        sid = setup_session(self.sdb, uid)
        # Now try to look it up
        sids_by_sub = self.sdb.get_sids_from_uid(uid)
        self.assertEqual(len(sids_by_sub), 1)
        self.assertIn(sid, sids_by_sub)

    def test_distributed_get_sids_from_uid(self):
        # This fails
        # Like the former test, but with multiple sdb instances
        # Sharing one db
        uid = 'uiduid'
        db = self.db
        # First
        sdb1 = SessionDB('https://1.com', db=db)
        sid1 = setup_session(sdb1, uid)
        # Second
        sdb2 = SessionDB('https://2.com', db=db)
        sid2 = setup_session(sdb2, uid)
        # Find both
[...]

PR #387, opened 2017-06-29
Refactor SessionDB constructor and use secure defaults

schlenk:
Refactor the SessionDB constructor to be less overloaded.

Looking into #326 i found the password and secret parameters are only used when a DefaultToken factory is used, so they do not really belong into the SessionDB class, if we externalize the factories. The seed parameter is not used at all.

So i threw out the parameters from the __init__ and wrote a wrapper that is nearly as convenient as the old API for the test/example cases.

This partially fixes https://github.com/OpenIDC/pyoidc/issues/326 as the defaults are gone, but DefaultToken does not raise nice exceptions yet.

This is still work-in-progress, just wanted to get some extra looks and agreement about the direction this goes.

It would still be nice to refactor a bit more (e.g. the RefreshDB/RefreshTokenFactory divide seems unnecessary) and to get to a point where #146 can be fixed, so one can do a persistent SessionDB without too much troubles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,10 @@ The format is based on the [KeepAChangeLog] project.
 - [#325]: `oic.oic.claims_match` implementation refactored.
 - [#368]: `oic.oauth2.Client.construct_AccessTokenRequest()` as well as `oic.oic.Client` are now able to perform proper Resource Owner Password Credentials Grant
 - [#374]: Made the to_jwe/from_jwe methods of Message accept list of keys value of parameter keys.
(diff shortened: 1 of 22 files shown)


rohe:
I support this. LGTM.
I'd love to rewrite the whole SessionDB but ...

lwm:
@schlenk, do you want to try and get this in for https://github.com/OpenIDC/pyoidc/pull/403?

schlenk:
Would be nice if it is in 0.11. It is a step in the right direction, even if it doesn't go all the way yet.

lwm:
I'll give it a review and try to merge now.

lwm:
:beers: :beers: :beers:

PR #388, opened 2017-06-29
Allow to set Cookie path and domain

### LINKED ISSUE #380: CookieDealers should be able to set domain/path for their cookies
The CookieDealer baseclasses `create_cookie()` method has no way to set a path or domain for their cookies.

It should be possible to set the cookie path and domain via the passed in srv object like it is for the cookie_name.

schlenk:
This fixes #380.


rohe:
LGTM

Issue #392, opened 2017-07-03
DefaultToken.key() is a bit too complex for what it does


schlenk:
The DefaultToken key() method generates a session ID that is used by the session DB for storage and lookup of the session data.

To do this it combines various more or less random data plus some session dependend data and calculates a HMAC(SHA224) over it. 

But the HMAC is never verified! So why use a HMAC at all, the only point is that you can verify it.

The key() method of DefaultToken could be simplified to:

    h = hashlib.new('sha224')
    h.update(os.urandom(16))
    return h.hexdigest()

This would have the same security properties.

In addition it is not really clear why the method is part of the DefaultToken class at all, it would be more appropriate in the SessionDB class, as the \_\_call\_\_ method of DefaultToken takes a sid as parameter.

PR #393, opened 2017-07-03
Simplify sid construction

schlenk:
SHA224 is as good as HMAC-SHA224 when no validation is performed.

See Issue #392.


schlenk:
coveralls is silly: deleted 24 lines, coverage dropped by 18 lines

lwm:
:fire:

PR #401, opened 2017-07-27
Decode and verify

rohe:

---

Working with the OP test tools a misbehaviour in pyoidc became obvious.
In protocol messages the value None on a parameter means that that parameter should just be ignored.
This is not true for Claims and IDToken's. 
None values on parameters in either of these messages should be regarded as errors.

There also where some transformation errors when creating Message instances.
Most notably a non-integer was accepted as a value to a parameter marked as having integer value.

And lastly a problem with Python was discovered.
For some reason isinstance(True, int) == True
So, I had to add some code to work around this.

### DIFF
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -431,6 +431,12 @@ class Message(MutableMapping):
         else:
             if val is None:
                 self._dict[skey] = None
+            elif isinstance(val, bool):
+                if vtyp is bool:
+                    self._dict[skey] = val
+                else:
+                    raise ValueError(
+                        '"{}", wrong type of value for "{}"'.format(val, skey))
             elif isinstance(val, vtyp):  # Not necessary to do anything
                 self._dict[skey] = val
             else:
@@ -439,6 +445,15 @@ class Message(MutableMapping):
                         val = _deser(val, sformat="dict")
                     except Exception as exc:
                         raise DecodeError(ERRTXT % (key, exc))
+                elif vtyp == int:
+                    try:
+                        self._dict[skey] = int(val)
+                    except ValueError:
+                        raise ValueError(
+                            '"{}", wrong type of value for "{}"'.format(val,
+                                                                        skey))
+                    else:
+                        return
 
                 if isinstance(val, six.string_types):
                     self._dict[skey] = val
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -476,11 +476,12 @@ class Message(MutableMapping):
     def _add_key(self, keyjar, issuer, key, key_type='', kid='',
                  no_kid_issuer=None):
 
-        try:
-            logger.debug('Key set summary for {}: {}'.format(
-                issuer, key_summary(keyjar, issuer)))
-        except KeyError:
+        if issuer not in keyjar:
             logger.error('Issuer "{}" not in keyjar'.format(issuer))
+            return
+
+        logger.debug('Key set summary for {}: {}'.format(
+            issuer, key_summary(keyjar, issuer)))
 
         if kid:
             _key = keyjar.get_key_by_kid(kid, issuer)
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -445,7 +445,7 @@ class Message(MutableMapping):
                         val = _deser(val, sformat="dict")
                     except Exception as exc:
                         raise DecodeError(ERRTXT % (key, exc))
-                elif vtyp == int:
+                elif vtyp is int:
                     try:
                         self._dict[skey] = int(val)
                     except (ValueError, TypeError):
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -448,7 +448,7 @@ class Message(MutableMapping):
                 elif vtyp == int:
                     try:
                         self._dict[skey] = int(val)
-                    except ValueError:
+                    except (ValueError, TypeError):
                         raise ValueError(
                             '"{}", wrong type of value for "{}"'.format(val,
                                                                         skey))
--- a/src/oic/oic/message.py
+++ b/src/oic/oic/message.py
@@ -546,10 +546,9 @@ class OpenIDSchema(Message):
                     except ValueError:
                         raise VerificationError("Birthdate format error", self)
 
-        for val in self.values():
-            if val is None:
-                return False
-            
+        if any(val is None for val in self.values()):
+            return False
+
         return True
 
 
--- a/src/oic/oic/message.py
+++ b/src/oic/oic/message.py
@@ -546,6 +546,10 @@ class OpenIDSchema(Message):
                     except ValueError:
                         raise VerificationError("Birthdate format error", self)
 
+        for val in self.values():
+            if val is None:
+                return False
+            
         return True
 
 
--- a/src/oic/utils/keyio.py
+++ b/src/oic/utils/keyio.py
@@ -333,6 +333,16 @@ class KeyBundle(object):
         self._uptodate()
         return [key.kid for key in self._keys if key.kid != ""]
 
+    def remove_outdated(self, after):
+        now = time.time()
+        _kl = []
+        for k in self._keys:
+            if k.inactive_since and k.inactive_since + after < now:
+                pass
+            else:
+                _kl.append(k)
+        self._keys = _kl
+
 
 def keybundle_from_local_file(filename, typ, usage):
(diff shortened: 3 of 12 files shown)


src/oic/oic/message.py
```diff
@@ -546,6 +546,10 @@ def verify(self, **kwargs):
                     except ValueError:
                         raise VerificationError("Birthdate format error", self)
 
+        for val in self.values():
```
lwm: Could be `if any(val is None for val in self.values()):`

src/oic/utils/keyio.py
```diff
@@ -333,6 +333,16 @@ def kids(self):
         self._uptodate()
         return [key.kid for key in self._keys if key.kid != ""]
 
+    def remove_outdated(self, after):
```
lwm: Feel like adding a docstring?

src/oic/utils/keyio.py
```diff
@@ -333,6 +333,16 @@ def kids(self):
...
+        now = time.time()
+        _kl = []
+        for k in self._keys:
```
lwm: With my 'pythonic' hat on:

``` python
try:
    if not k.inactive_since + after < now:
        _kl.append(k)
except WhateverErrorThatIs:
    pass
```

src/oic/utils/keyio.py
```diff
@@ -396,6 +407,7 @@ def __init__(self, ca_certs=None, verify_ssl=True, keybundle_cls=KeyBundle):
         self.ca_certs = ca_certs
         self.verify_ssl = verify_ssl
         self.keybundle_cls = keybundle_cls
+        self.remove_after = remove_stale_after
```
lwm: The naming is different, by choice or? The rest are the same.

src/oic/utils/keyio.py
```diff
@@ -777,6 +789,39 @@ def get_issuer_keys(self, issuer):
...
+
+        # The set of issuers MUST be the same
+        if set(self.keys()) != set(other.keys()):
+            return False
+
+        # Keys per issuer must be the same
+        for iss in self.keys():
+            sk = self.get_issuer_keys(iss)
+            ok = other.get_issuer_keys(iss)
+            if len(sk) != len(ok):
```
lwm: The comment above suggests the key *values* must be the same, but only length or?

src/oic/utils/keyio.py
```diff
@@ -777,6 +789,39 @@ def get_issuer_keys(self, issuer):
...
+                return False
+
+            for k in sk:
```
lwm: Could be: `if not any(k in ok for k in sk):`

src/oic/utils/keyio.py
```diff
@@ -777,6 +789,39 @@ def get_issuer_keys(self, issuer):
...
+                if k not in ok:
+                    return False
+
+        return True
+
+    def remove_outdated(self):
```
lwm: Oh, another method with the same name? Now I demand a docstring :)

src/oic/utils/keyio.py
```diff
@@ -777,6 +789,39 @@ def get_issuer_keys(self, issuer):
...
+        for iss in list(self.keys()):
```
lwm: Is the `list()` wrapper necessary?

src/oic/oauth2/message.py
```diff
@@ -431,6 +431,12 @@ def _add_value(self, skey, vtyp, key, val, _deser, null_allowed):
         else:
             if val is None:
                 self._dict[skey] = None
+            elif isinstance(val, bool):
+                if vtyp is bool:
```
lwm: Would prefer to see `if isinstance(vtyp, bool)`.

src/oic/oauth2/message.py
```diff
@@ -439,6 +445,15 @@ def _add_value(self, skey, vtyp, key, val, _deser, null_allowed):
                         val = _deser(val, sformat="dict")
                     except Exception as exc:
                         raise DecodeError(ERRTXT % (key, exc))
+                elif vtyp == int:
```
lwm: Again, would prefer an `isinstance` call.

src/oic/oauth2/message.py
```diff
@@ -439,6 +445,15 @@ def _add_value(self, skey, vtyp, key, val, _deser, null_allowed):
...
+                    try:
+                        self._dict[skey] = int(val)
+                    except (ValueError, TypeError):
```
lwm: I don't see why we should ignore `TypeError` here? If your only wish is to pass a message with more information you can do `except (ValueError, TypeError) as exc:  raise exc(msg)`.

src/oic/oic/message.py
rohe: OK

src/oic/utils/keyio.py
rohe: Always ...

src/oic/utils/keyio.py
rohe: Can buy that

src/oic/utils/keyio.py
rohe: By choice but I guess mostly of historic reasons. 
So should possibly make them the same.

src/oic/utils/keyio.py
rohe: The set of keys must be that same which means that the number of keys must be the same.
That's a MUST but not sufficient as the lines below show.
It's an optimisation. If the number of keys are different then I don't have to check the
individual keys.

src/oic/utils/keyio.py
rohe: Could ..

src/oic/utils/keyio.py
rohe: Sure, different classes so slightly different even though the basic task is the same.

src/oic/utils/keyio.py
rohe: Probably not. But I've been bitten by not using it so many times now that it's become a default.

src/oic/oauth2/message.py
rohe: Hmm

src/oic/oauth2/message.py
rohe: I'm not ignoring TypeError. I'm just raising a ValueError if for some reason 'val' can not be 
cast to an int.

rohe:
Thanks for the serious review Luke! Much appreciated !

src/oic/oauth2/message.py
rohe: vtyp is not an instance of bool. It is the type bool so doing an isinstance test won't work.

src/oic/oauth2/message.py
rohe: Same as above, vtyp is a type not an instance of a type.

rohe:
@lwm if you're OK with the changes I made please merge till PR.
Time to release a new (minor) version ?

lwm:
> Time to release a new (minor) version ?

If we make a next release, [it will need to be major](https://github.com/OpenIDC/pyoidc/blob/master/CHANGELOG.md#01100-unreleased). Yes, it's probably a good idea to do this now.

tests/test_keyio.py
```diff
@@ -430,3 +431,44 @@ def test_get_signing_key_use_undefined():
...
+    assert len(_interm) == 4
+
+    # Now mark the keys to be inactivated
+    _now = time.time()
+    for k in keyjar.get_issuer_keys(''):
+        if k.kid in _old:
+            if not k.inactive_since:
+                k.inactive_since = _now
+
+    time.sleep(2)
```
lwm: Oh! This is generally a bad idea. The slower the tests get, the more likely we are not to run them. I would absolutely recommend using [freezegun](https://github.com/spulec/freezegun) and a [manual tick](https://github.com/spulec/freezegun#manual-ticks) (which bumps time along instantly!).

schlenk:
We could always still do a minor release. Its just the number in the changelog... (and next time we should probably not set any version number in the changelog for unreleased until we decide on the final number and release it).

lwm:
> We could always still do a minor release.
> next time we should probably not set any version number in the changelog for unreleased

I think we have breaking changes included in the latest pull requests though? I think we named the new version number because there was an incompatible change, so it had to be this major bump. Could be wrong ...

schlenk:
yes, probably right.

tests/test_keyio.py
lwm: @rohe, I'll add this change the to PR now and merge soon after.

lwm:
:beers: :beers: :beers:

PR #406, opened 2017-08-10
Fix generation of endpoint urls

### LINKED ISSUE #405: Incorrectly created endpoint urls
As mentioned in #375, `urljoin` does strip parts of path in some cases and produced incorrect endpoint URLs.

The current implementation has also some flaws...

The main issue is that if the `base_url` doesn't end with a slash, it is stripped by `urljoin`.

If the `base_url` ends with a slash, the currently produced URl is wrong.

tpazderka:
Close #405


lwm:
Ah yes, nice catch. Looks solid to me, add a change log entry and merge as you see fit :+1:

Issue #407, opened 2017-08-11
phone_number_verified is a boolean claim


zandbelt:
https://github.com/OpenIDC/pyoidc/blob/master/src/oic/oic/message.py#L527

marks `phone_number_verified` as a string whereas the spec says it is a boolean http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

lwm:
PRs appreciated @zandbelt! Otherwise, let's hope someone gets to this soon.

rohe:
This is fixed by https://github.com/OpenIDC/pyoidc/pull/418.

PR #408, opened 2017-08-11
make phone_number_verified a boolean in OpenIDSchema

zandbelt:
addresses OpenIDC/pyoidc#407

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,10 @@ The format is based on the [KeepAChangeLog] project.
 
 ### Fixed
 - [#405]: Fix generation of endpoint urls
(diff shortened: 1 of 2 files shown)


lwm:
That's awesome that you put time on this one @zandbelt! Could I ask you to write a test for this? It might seem really boring but one line changes always scare me. Especially when the other tests pass ;)

rohe:
I definitely support this change, it's really fixing a bug.
But @lwm is right we should add a test.

rohe:
Wrote a couple of tests. They showed a bug in the code. Will submit a PR.

lwm:
Thanks @rohe, will close off this PR for #418 (this change set merged there). 

Thanks for contributing @zandbelt!

PR #414, opened 2017-08-24
Correctly create error response when wrong response_mode requested

### LINKED ISSUE #413: Server error if wrong `response_mode` is requested
When consumer asks for a specific `response_mode` and this cannot be fulfilled, an error response is produced, but not correctly handled.
```
Exception Type: TypeError at /oidc/authorization/
--
Exception Value: InvalidRequest('wrong response_mode',) is not JSON seriali=
zable
```
This should be correctly handled and a proper error response should be returned.

tpazderka:
Close #413


lwm:
LGTM!

Issue #419, opened 2017-08-30
Inconsistent release numbers/tags


tpazderka:
For the latest release (I haven't checked any others), there is an inconsistency in version number and tag.
Tag is `v0.11.1`, version in changelog is the same as well as version on github project page, but version in code and in PIP is `0.11.0.1`.

We should probably make them the same.

lwm:
Ah, I see. I've done this before ;) I'll take a look soon, thanks for the report!

lwm:
Ah, taking another look at this, it isn't my fault :) 

So, AFAICT, the code always used this `X.X.X.X` format (and I wasn't quite sure why). The tags always followed the `vX.X.X` format. All of this existed before I started co-maintaining the project. @rohe, can you clear this up this difference? Was it intentional?

> Tag is v0.11.1, version in changelog is the same

Actually, the change log conforms to the `X.X.X.X` format.

So, the fix here would be to recreate all the tags using the `X.X.X.X` format. That would be some work.

tpazderka:
I think that the main problem is that release on Github, PIP and version in the code are not the same.

After we decide on one format, we probably can change only the last two releases?

schlenk:
I wouldn't change the past, especially not released packages, that creates more harm than it does good. Maybe create additional tags in the repo that match what was released, for easier diffing, but even that is IMHO optional.

We should simply commit to some future format (x.x.x.x seems common for python these days) and keep that.

rohe:
I don't think I actually had a thought through reason for using x.x.x.x . It was probably along the line @schlenk points to. Namely that it is what is commonly used.
I don't have a strong preference for x.x.x or x.x.x.x .
So, what ever the majority (hey we are a even number of collaborators so we may never resolve this :-)) decides is OK with me.
Having said that, why not continue with x.x.x.x  ?
And I agree with @tpazderka I see no reason going beyond the last two releases.

rohe:
Along the same lines. We should have an agreement on what constitutes the difference between a moderate and a minor release.
And I don't think I've ever used the 4th digit. So in reality I've used x.x.x rather then x.x.x.x .

tpazderka:
I am not against either of those formats. My point is to keep that consistent throughout the project :)

Minor release should be a bug fixes and such. If there are improvements, added functionality or incompatibility, I would consider it a moderate change.

rohe:
[...]

PR #420, opened 2017-09-04
Distributed claims

rohe:
Fixed bug in handling of distributed claims.

---

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -764,6 +764,7 @@ class Client(oauth2.Client):
     def user_info_request(self, method="GET", state="", scope="", **kwargs):
         uir = UserInfoRequest()
         logger.debug("[user_info_request]: kwargs:%s" % (sanitize(kwargs),))
+        token = None
         if "token" in kwargs:
             if kwargs["token"]:
                 uir["access_token"] = kwargs["token"]
@@ -773,12 +774,11 @@ class Client(oauth2.Client):
                 kwargs["behavior"] = "use_authorization_header"
             else:
                 # What to do ? Need a callback
-                token = None
+                pass
         elif "access_token" in kwargs and kwargs["access_token"]:
             uir["access_token"] = kwargs["access_token"]
             del kwargs["access_token"]
-            token = None
-        else:
+        elif state:
             token = self.grant[state].get_token(scope)
 
             if token.is_valid():
@@ -808,17 +808,19 @@ class Client(oauth2.Client):
         if "behavior" in kwargs:
             _behav = kwargs["behavior"]
             _token = uir["access_token"]
+            _ttype = ''
             try:
                 _ttype = kwargs["token_type"]
             except KeyError:
-                try:
-                    _ttype = token.token_type
-                except AttributeError:
-                    raise MissingParameter("Unspecified token type")
+                if token:
+                    try:
+                        _ttype = token.token_type
+                    except AttributeError:
+                        raise MissingParameter("Unspecified token type")
 
             if 'as_query_parameter' == _behav:
                 method = 'GET'
-            else:
+            elif token:
                 # use_authorization_header, token_in_message_body
                 if "use_authorization_header" in _behav:
                     token_header = "{type} {token}".format(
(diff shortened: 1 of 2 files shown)


src/oic/oic/__init__.py
```diff
@@ -1064,12 +1066,16 @@ def fetch_distributed_claims(self, userinfo, callback=None):
...
                 if "access_token" in spec:
                     _uinfo = self.do_user_info_request(
-                        token=spec["access_token"],
+                        method='GET', token=spec["access_token"],
                         userinfo_endpoint=spec["endpoint"])
                 else:
-                    _uinfo = self.do_user_info_request(
-                        token=callback(csrc),
-                        userinfo_endpoint=spec["endpoint"])
+                    if callback:
```
tpazderka: This part is not covered by tests.

src/oic/oic/__init__.py
```diff
@@ -808,17 +808,19 @@ def user_info_request(self, method="GET", state="", scope="", **kwargs):
...
+                if token:
+                    try:
+                        _ttype = token.token_type
+                    except AttributeError:
+                        raise MissingParameter("Unspecified token type")
 
             if 'as_query_parameter' == _behav:
                 method = 'GET'
-            else:
+            elif token:
```
tpazderka: * Comparison to `None` should be explicit.
* In tests, this condition is always true, so a test with token = `None` should be added.

src/oic/oic/__init__.py
rohe: Does the recent changes meet your requirements ?

tpazderka: Sorry for not responding earlier. I am having a busy week at work. I will have a look today.

src/oic/oic/__init__.py
tpazderka: The comparison is still not explicit. It believe that it should be `elif token is not None:` as token is previously assigned `None`.

src/oic/oic/__init__.py
```diff
@@ -808,17 +808,19 @@ def user_info_request(self, method="GET", state="", scope="", **kwargs):
...
             _token = uir["access_token"]
+            _ttype = ''
             try:
                 _ttype = kwargs["token_type"]
             except KeyError:
-                try:
-                    _ttype = token.token_type
-                except AttributeError:
-                    raise MissingParameter("Unspecified token type")
+                if token:
```
tpazderka: The same goes here: `if token is not None:`

src/oic/oic/__init__.py
tpazderka: Here again, comparison to `None` and test without callback is missing.

src/oic/oic/__init__.py
```diff
@@ -899,12 +901,13 @@ def do_user_info_request(self, method="POST", state="", scope="openid",
...
-        # Verify userinfo sub claim against what's returned in the ID Token
-        idt = self.grant[state].get_id_token()
-        if idt:
-            if idt['sub'] != res['sub']:
-                raise SubMismatch(
-                    'Sub identifier not the same in userinfo and Id Token')
+        if state:
+            # Verify userinfo sub claim against what's returned in the ID Token
+            idt = self.grant[state].get_id_token()
+            if idt:
```
tpazderka: This seems to never be triggered in tests.

tpazderka:
I know this is a lot of tests to be added, but if you are already poking into the code, we might as well add test to cover the changes...

src/oic/oic/__init__.py
rohe: I don't think that is good enough. token='' should be handled the same way as token=None
hence the use of ´<elif token>´ which implicitly include <token is not ''>.
So I'd rather not use not None since then I would have to have do
<elif token is not None and token is not ''>

tpazderka: And is there any case where token would be an empty string?
If yes, then OK, leave the code as is.

src/oic/oic/__init__.py
rohe: OK now ? Need to get this out the door soon.

tpazderka:
I would like to see more tests, but I understand that any progress is better than none.

lwm:
> I would like to see more tests, but I understand that any progress is better than none.

Agreed!

If we feel like we're missing coverage, just create an issue with a link to this PR and we can track that. If a bug is ever raised, we will be motivated to come back to cover the remaining cases.

Nice work @rohe.

rohe:
Sounds good to me.
So, can one of you merge the PR.
Since I'm the author I shouldn't do it myself.
After the merge I'd like us to release a new version to pypi. 
The next release of the OpenID Connect tool has a dependency on this PR and I want to get that out the door as soon as possible.

lwm:
🍰 

> Since I'm the author I shouldn't do it myself.

Hey, you may be VIP but you're not getting out of merging your own PRs ;)

I won't get time to make a release, please someone, go for it and keep https://github.com/OpenIDC/pyoidc/issues/419 in mind.

rohe:
Excellent new! Thanks !
VIP, lol :-)
I'll probably not have time to do a release today. At the earliest tomorrow afternoon.

PR #426, opened 2017-09-25
Release 0.12.0

rohe:
From now on we will use X.X.X as structure for release numbers and tags.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ The format is based on the [KeepAChangeLog] project.
 
 [KeepAChangeLog]: http://keepachangelog.com/
 
+## 0.12.0 [2017-09-25]
+
+### Fixed
+- [#419]: Inconsistent release numbers/tags
+- [#420]: Distributed claims
+
+[#419]: https://github.com/OpenIDC/pyoidc/issues/419
+[#420]: https://github.com/OpenIDC/pyoidc/pull/420
+
 ## 0.11.1.0 [2017-08-26]
 
 ### Fixed
--- a/src/oic/__init__.py
+++ b/src/oic/__init__.py
@@ -8,7 +8,7 @@ except ImportError:
     import random as rnd
 
 __author__ = 'Roland Hedberg'
-__version__ = '0.11.0.1'
+__version__ = '0.12.0'
 
 
 OIDCONF_PATTERN = "%s/.well-known/openid-configuration"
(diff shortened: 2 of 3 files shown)


rohe:
I don't know why the problem with hmac.compare_digest appears now and not before.
Should we do something similar to https://github.com/carlos-jenkins/python-github-webhooks/pull/5/commits/c225d20d715a870c0f369ae0d3d2d89c25a3f091 ?

tpazderka:
The issue is visible on `master` which has failing builds.

I am not sure if we can limit TOX to run on python 2.7.7 but if not, then we should revert to the old behaviour for 2.7?

rohe:
I also don't know if we can limit TOX to use Python 2.7.7 . 
To me it looks like you only can specify 2.6, 2.7 and so on. Not 2.7.6, 2.7.7, ...

I'd like Michael's view on this. 
He added (for good reasons) constant time verifier for signatures and I'm sure has views on how we should handle Python < 2.7.7.

schlenk:
For Python < 2.7.7 the hmac.compare_digest() code doesn't exist (sadly).

As we depend on a rather fresh cryptography package, we could simply use the constant_time compare function used there:
`cryptography.hazmat.primitives.constant_time.bytes_eq`
https://cryptography.io/en/latest/hazmat/primitives/constant-time/

It didn't use it, because hmac looked easier and handles both bytes/strings, while the above needs bytes explictly.

schlenk:
Seems Travis randomly picks Python 2.7.6 or 2.7.13 for the tests. 15 days ago in https://travis-ci.org/OpenIDC/pyoidc/builds/273213747 it picked 2.7.6 and things broke, the next build it picked 2.7.13 again and it worked, and since a few days it always seems to pick 2.7.6.

schlenk:
Found an even better idea (as my assumption about cryptography.io being required isn't true if the pyopenssl is old too): https://github.com/rohe/pyjwkest/blob/master/src/jwkest/__init__.py has a constant time string compare, so lets use that one.

tpazderka:
@schlenk Good idea!

rohe:
OK, so we'll drop hmac.compare_digest() and instead use constant_time_compare() from pyjwkest.
Is that correct ?

schlenk:
For now, yes. 
hmac.compare_digest() is faster of course, but to get the release out of the door i would suggest to do the easy thing.

Issue #424, opened 2017-09-21
userinfo_endpoint return data if access_token is expired


ermakoves:
Is this normal behavior?

lwm:
That sounds like a bug to me.

Anyone else?

schlenk:
Seems like oic.provider _do_user_info() should check for token expiry, it only checks for revocation and existance in the sdb.

rohe:
Agree, so who makes the PR ?
Any taker.

Issue #328, opened 2017-04-12
`grant_expires_in` is not honoured


lwm:
The `SessionDB` takes a `grant_expires_in` kwarg which doesn't seem to be working.

I had to patch the `oic.oic.provider.token_endpoint` with the following:

``` python
    def token_endpoint(self, request="", **kwargs):
        data = dict(parse_qsl(request))

        try:
            event = self.sdb[data['code']]['authn_event']
        except KeyError:
            return Response(status="400 Bad Request")

        if int(time()) > event.valid_until:
            return Response(status="498 Invalid Token")

        return super().token_endpoint(request, **kwargs)
```

I don't see any tests covering this, so I think this might be an oversight?

lwm:
Probably `int(time()) > event.valid_until:` is not the right check ...

It should probably be `int(time()) > event.authn_time + self.sdb.token_factory['code'].lifetime`

rohe:
Sorry, I've been otherwise occupied for a couple of days.

Time is extremely confusing. 
And you're right in that lifetime should have been added.

tpazderka:
This has been fixed in #429

PR #427, opened 2017-09-27
Response types

rohe:
Dealing with issue #399 

Sorry about the extra commits. Just tried to even them out.

Thanks for contributing to this library! Please include the following check
list in your pull request submission (you can delete this message). If your
changes don't need a change log or documentation update, please ignore this.

- [X ] Any changes relevant to users are recorded in the `CHANGELOG.md`.
- [ ] The documentation has been updated, if necessary.

---

PR #428, opened 2017-09-27
WIP: Add expiration time checker.

lwm:
I started out trying to fix https://github.com/OpenIDC/pyoidc/issues/424, and then remembered I had previously raised https://github.com/OpenIDC/pyoidc/issues/328. So, this is a stab at adding a reliable way to check if the access token is expired, which will help close both. As mentioned in https://github.com/OpenIDC/pyoidc/issues/328#issuecomment-293566067, and still, I'm a little confused as to whether this is the right approach. Reviews very welcome!

Tests are failing right now. Will fix once I'm more sure.

---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -1182,6 +1182,7 @@ class Provider(AProvider):
         except ParameterError:
             return error(error='invalid_request',
                          descr='Token is malformed')
+
         return self._do_user_info(_token, **kwargs)
 
     def _parse_access_token(self, request, **kwargs):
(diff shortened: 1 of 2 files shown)


schlenk:
I think the details should be a task for the token_factory, not the SDB. 

This would also make it easier to also handle access code expiry and refresh token expiry.

The time-of-check should probably be an optional parameter (and default to time.time() if not given/None). Makes testing easier too :-).

You should use time.time(). Thats always POSIX epoche seconds in UTC timezone.
Do NOT use datetime.datetime.now() as that is always in localtime (e.g. affected by timezones and makes weird stuff if you load balance servers in different timezones).

rohe:
I also started to look at this problem. Had a PR in the making just waiting for the other one I have active now to come through.

Agree with @schlenk this should be handled by a method in the token_factory.
My solution was based on a is_expired method in the sdb.Token class.

Good idea to have time-of-check as an optional parameter, hadn't though about that.
I used sleep in the test which is a bit awkward (time consuming).

lwm:
Awesome, thanks! Should I close this off awaiting your change set @rohe? Otherwise Ill go with recommendations.

rohe:
As soon as #427 has gone through I can push my PR.
We should start using assignees so we're not unnecessarily working on the same issues.

lwm:
> We should start using assignees so we're not unnecessarily working on the same issues.

Agreed, I've assigned you on those issues. I'll close this off and grab something else 👍

PR #430, opened 2017-10-06
Audience is endpoint dependent.

rohe:

---

### DIFF
--- a/src/oic/utils/authn/client.py
+++ b/src/oic/utils/authn/client.py
@@ -45,6 +45,7 @@ def assertion_jwt(cli, keys, audience, algorithm, lifetime=600):
     at = AuthnToken(iss=cli.client_id, sub=cli.client_id,
                     aud=audience, jti=rndstr(32),
                     exp=_now + lifetime, iat=_now)
+    logger.debug('AuthnToken: {}'.format(at.to_dict()))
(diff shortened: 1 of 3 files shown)


rohe:
I really don't care if the coverage decreases with 0.001 % .

src/oic/utils/authn/client.py
```diff
@@ -310,14 +311,16 @@ def construct(self, cis, request_args=None, http_args=None, **kwargs):
...
-        audience = self.cli.provider_info['issuer']
         algorithm = None
         if kwargs['authn_endpoint'] in ['token', 'refresh']:
             try:
                 algorithm = self.cli.registration_info[
                     'token_endpoint_auth_signing_alg']
             except (KeyError, AttributeError):
                 pass
+            audience = self.cli.provider_info['token_endpoint']
+        else:
```
tpazderka: The drop in coverage is caused by this branch, which is never covered.

rohe: Can easily be remedied

rohe:
I don't think this will break anything. It's really a corner case, slightly outside the standard.
private_key_jwt and cleint_secret_jwt was designed for client authentication at the token endpoint.
But now when the methods are defined they are creatively :-) used at other endpoints.
In which case using the token endpoint URL as audience doesn't make sense.

PR #436, opened 2017-11-03
Fix headers for read_registration

schlenk:
Headers was passed as list instead of as a dictionary.

---


lwm:
That syntax looks a bit dodge? I tried:

``` python
>>> {"Authorization", "Bearer %s" % "token"}['Authorization']
  File "<stdin>", line 1, in <module>
TypeError: 'set' object has no attribute '__getitem__'
```

It actually produces a `set`?

tpazderka:
Because there is a comma instead of a colon. Also tests covering this change are missing.

lwm:
Yep, but the first comment says @schlenk is expecting a dictionary.

Would be nice to add a test, for sure!

schlenk:
I'll have a look if i can come up with a good test for this too.

PR #444, opened 2017-12-01
Add ability to specify custom form_post template

### LINKED ISSUE #334: Improve form_post capabilities
Current support for `form_post` method uses predefined template and does not allow for its substitution and redefinition.

The predefined form itself contains an invalid HTML (missing `"` around `action`) and contains inline JS which is not allowed in Content Security Policy.

The included form should be a valid HTML and should not contain inline JS in order to be usable on sites with CSP enabled.

It should also be possible to replace the form with a custom one.

tpazderka:
---
Close #334

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -83,6 +83,7 @@ from oic.utils.keyio import key_export
 from oic.utils.sanitize import sanitize
 from oic.utils.sdb import AccessCodeUsed
 from oic.utils.sdb import ExpiredToken
+from oic.utils.template_render import render_template
 from oic.utils.time_util import utc_time_sans_frac
 
 __author__ = 'rohe0002'
@@ -208,28 +209,6 @@ CAPABILITIES = {
     "request_uri_parameter_supported": True,
 }
 
-FORM_POST = """<html>
-  <head>
-    <title>Submit This Form</title>
-  </head>
-  <body onload="javascript:document.forms[0].submit()">
-    <form method="post" action={action}>
-        {inputs}
-    </form>
-  </body>
-</html>"""
-
-
-def inputs(form_args):
-    """
-    Creates list of input elements
-    """
-    element = []
-    for name, value in form_args.items():
-        element.append(
-            '<input type="hidden" name="{}" value="{}"/>'.format(name, value))
-    return "\n".join(element)
-
 
 class Provider(AProvider):
     def __init__(self, name, sdb, cdb, authn_broker, userinfo, authz,
@@ -237,7 +216,7 @@ class Provider(AProvider):
                  hostname="", template_lookup=None, template=None,
                  verify_ssl=True, capabilities=None, schema=OpenIDSchema,
                  jwks_uri='', jwks_name='', baseurl=None, client_cert=None,
-                 extra_claims=None):
+                 extra_claims=None, template_renderer=render_template):
 
         AProvider.__init__(self, name, sdb, cdb, authn_broker, authz,
                            client_authn, symkey, urlmap,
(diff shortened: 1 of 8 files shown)


src/oic/oic/provider.py
```diff
@@ -234,7 +234,8 @@ def __init__(self, name, sdb, cdb, authn_broker, userinfo, authz,
                  client_authn, symkey=None, urlmap=None, ca_certs="", keyjar=None,
                  hostname="", template_lookup=None, template=None,
                  verify_ssl=True, capabilities=None, schema=OpenIDSchema,
-                 jwks_uri='', jwks_name='', baseurl=None, client_cert=None):
+                 jwks_uri='', jwks_name='', baseurl=None, client_cert=None,
+                 form_post_template=FORM_POST):
```
schlenk: A name that hints it is about the form_post response_mode would be better.

The provider already has two template related args (template_lookup/template). Why not reuse the template_lookup to lookup a custom form_post response template too?

tpazderka: Good catch. I can probably reuse the `template` arg.

I am not sure how (and for what) is the `template_lookup` used as its default value would cause error.

tpazderka:
OK, I cannot find any documentation about the `template` arg structure.

I guess that the structure depends on the value of `template_lookup` and in order to reuse this, it would require some work to implement a default `template_lookup` method.

Would you prefer the `template` to store raw templates as strings or paths to a file with template?

With that I can implement a default `template_lookup` function and go from there.

schlenk:
As far as i can see, the template/template_lookup thing is mostly used for the utils/http_util.py:Response code to render a template. And it is used rather inconsistently.

So template is a dict of 'template_names' (those should be documented somewhere), and some lookup key used by the template_lookup. The mako lookup uses relative filenames as far as i can see.

So i think the values should not be simple strings, but could be rather trivial function objects that have a very basic render method that mostly boils down to a `format()`.

If you look at the examples (op2), there seems to be a name collision with your proposed template name (maybe its the same thing).
oidc_example\op2\server.py
534:        "template": {"form_post": "form_response.mako"},

tpazderka:
OK, then.

I will try to implement this with default template_lookup as mako with directory somewhere in the project.

I will look into documenting the changes.

tpazderka:
Reworked a little bit.

I decided to remove `template_lookup` and `template` arguments (they look unused?) and replaced them with a `template_renderer` which should be capable of replacing them entirely.

The `verify_logout` template needs to be finished but I am not sure yet what was its function as it is nowhere to be found :) @rohe could you please explain the intended functions to me?

Documented and hopefully fixed the examples.

schlenk:
Looks good.

Not sure how the connection is to 'utils/http_utils.py:Response' which takes kwargs parameters for template_lookup/mako_template/template or falls back to some defaults. I didn't see right away how it might get there from the Provider class. Probably some templates are set for rendering the login forms etc.

tpazderka:
I will look into that. But if we figure out it is relevant and should be refactored, I suppose another issue would be a good place to keep the changes small and manageable.

schlenk:
ah well, seems some tests still fail, so fix those first, but the issue seems trivial.

tpazderka:
Yeah a typo :) And I am still waiting for the `verify_logout` template.

tpazderka:
OK, I have added some basic `verify_logout` template as I want to get this merged. We can improve upon it later if there is need for that.

schlenk:
OK. Feel free to merge.

The template rendering could probably use some hardening against malicious inputs later, but as far as i can see the input isn't really user controlled, so should be good enough for now. One could also move the onbody handler into an external file to help with Content-Security-Policies.

tpazderka:
I had CSP in mind but I think we cannot make any assumptions about deployment so hardcoding an external file might be difficult. This at least allows to override the template completely (styling, CSP and such).

PR #447, opened 2017-12-04
Fix KeyError on provider.read_registration

### LINKED ISSUE #446: Exception at read_registration on wrong authentication
When one performs a request on read_registration endpoint with wrong authentication info, uncaught exception is raised (https://github.com/OpenIDC/pyoidc/blob/master/src/oic/oic/provider.py#L1637).

The KeyError should be expected and server should return `error_response`.

Also `assert` on line 1641 should be replaced by some other checking mechanism because of optimalizations.

tpazderka:
Close #446

---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -1634,11 +1634,15 @@ class Provider(AProvider):
             return error_response('invalid_request')
         token = authn[len("Bearer "):]
 
-        client_id = self.cdb[token]
+        try:
+            client_id = self.cdb[token]
+        except KeyError:
+            return error_response('invalid_request')
 
         # extra check
         _info = parse_qs(request)
-        assert _info["client_id"][0] == client_id
+        if not _info["client_id"][0] == client_id:
+            return error_response('invalid_request')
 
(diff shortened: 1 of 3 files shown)


src/oic/oic/provider.py
```diff
@@ -1634,11 +1634,15 @@ def read_registration(self, authn, request, **kwargs):
...
+        try:
+            client_id = self.cdb[token]
+        except KeyError:
+            return error_response('invalid_request')
 
         # extra check
         _info = parse_qs(request)
-        assert _info["client_id"][0] == client_id
+        if not _info["client_id"][0] == client_id:
+            return error_response('invalid_request')
```
schlenk: https://github.com/OpenIDC/pyoidc/blob/master/src/oic/extension/provider.py#L466 has similar code (for a more fully featured registration info), and raises Unauthorized() here.

I think Unauthorized is more appropriate for this, as the Bearer token is wrong/non-existing or the client_id is wrong.

tpazderka: Changed both of them.

Issue #404, opened 2017-08-06
Stop coveralls spamming the issues


lwm:
Whenever I'm hacking on a PR, the multiple CI runs generate coveralls reports and then the bot starts spamming the pull request with comments. It'd be nice to reduce this in some way. Creating this to track that.

lwm:
According to https://github.com/lemurheavy/coveralls-public/issues/93#issuecomment-25119377, the owner of the coveralls account which has control over this repository can stop comments altogether. I think I'd be in favour of that - anyone against? If not, can someone do it? Otherwise, feel free to close this off - they don't seem to support only commenting once.

tpazderka:
This is rather unfortunate since we wouldn't have any visual control over the change in coverage...

But I do understand your frustration :)

PR #451, opened 2017-12-07
Use Codecov instead of coveralls

tpazderka:
---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,7 @@ install:
   - pip install tox
 
 script:
-  - tox -e coverage-erase,$TOX_ENV
-
-after_success:
-  - tox -e coveralls
+  - tox -e $TOX_ENV
 
 notifications:
   email: false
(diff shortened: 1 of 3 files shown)


tpazderka:
@rohe Can you please login to Codecov and try to enable either the Webhook or the Plugin?

We can try both and see what suits us better.

lwm:
Nice one!

tpazderka:
If that wasn't you Roland, then there is probably nothing else to do. It seems that Codecov reuses the comment and is not creating a new every time. I will try that tomorrow with couple of rebases...

tpazderka:
It looks like it works! :)

tpazderka:
OK guys, I think that this one looks much better than coveralls. The comment is reused so no spamming and the report is also more useful as it shows coverage change for changed files.

So if noone has any objections, I would like to merge this.

lwm:
go go go go go go go go go go go go

lwm:
For discovery, this deals with what we discussed in https://github.com/OpenIDC/pyoidc/issues/404.

tpazderka:
Yes :)

PR #452, opened 2017-12-08
Add travis builds under python 3.5

tpazderka:
Reorganized `.travis.yaml` to use matrix. This makes it possible to run tests under python 3.5.


lwm:
🏳️‍🌈

PR #456, opened 2017-12-11
Support for configuration endpoint

### LINKED ISSUE #134: Support Client Configuration Endpoint
http://openid.net/specs/openid-connect-registration-1_0.html#ClientConfigurationEndpoint

oic.oic.Provider:registration_endpoint works great, and the ClientRegistrationResponse includes the 'registration_client_uri'. However, if I issue GET requests to that URI, I usually get a 400 Bad Request response code. It looks like this is because the GET request is treated the same as a (new) Client Registration Request.

@rohe Is this "Client Read Request" functionality something intentionally omitted? Or would you welcome contributions to implement it?

tpazderka:
Close #134

---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -17,6 +17,7 @@ import socket
 import sys
 import time
 import traceback
+import warnings
 from functools import cmp_to_key
 
 import six
@@ -1512,6 +1513,11 @@ class Provider(AProvider):
             args[param] = val
 
     def l_registration_endpoint(self, request, authn=None, **kwargs):
+        warnings.warn('`l_registration_endpoint` is deprecated. Please use `create_registration` instead',
+                      DeprecationWarning)
+        return self.create_registration(request=request, authn=authn, **kwargs)
+
+    def create_registration(self, authn=None, request=None, **kwargs):
         logger.debug("@registration_endpoint: <<%s>>" % sanitize(request))
 
         try:
@@ -1611,7 +1617,7 @@ class Provider(AProvider):
 
     def registration_endpoint(self, request, authn=None, method='POST', **kwargs):
         if method.lower() == 'post':
-            return self.l_registration_endpoint(request, authn, **kwargs)
+            return self.create_registration(authn, request, **kwargs)
         elif method.lower() == 'get':
             return self.read_registration(authn, request, **kwargs)
         elif method.lower() == 'put':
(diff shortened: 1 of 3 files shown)


src/oic/oic/provider.py
```diff
@@ -1609,8 +1609,16 @@ def client_registration_setup(self, request):
 
         return response
 
-    def registration_endpoint(self, request, authn=None, **kwargs):
-        return self.l_registration_endpoint(request, authn, **kwargs)
+    def registration_endpoint(self, request, authn=None, method='POST', **kwargs):
+        if method.lower() == 'post':
+            return self.l_registration_endpoint(request, authn, **kwargs)
```
lwm: The name of `l_registration_endpoint` is quite out of place now - feel like updating that?

lwm:
Optional comment to deal with but otherwise, this looks good!

src/oic/oic/provider.py
tpazderka: Was thinking about it :) Unfortunately it is not marked as private so someone might be using it.

Also the order of args is not consistent with the `read_registration` so if we decide to change it, I would change both name and order of args.

lwm: OK, makes sense!

src/oic/oic/provider.py
```diff
@@ -1654,6 +1662,24 @@ def read_registration(self, authn, request, **kwargs):
...
                         headers=[("Cache-Control", "no-store")])
 
+    def alter_registration(self, authn, request, **kwargs):
+        """Method to alter the client info on server side.
+
+        :param authn: Authorization HTTP header
+        :param request: Query part of the request
+        :return: Response with updated client info
+        """
+        return error_response('Unsupported operation', descr='Altering of the registration is not supported')
```
schlenk: I would suggest a 403 response, in line with the RFC 7592 2.2 section, as that spec is likely the one someone would implement here. It demands a 403.

src/oic/oic/provider.py
```diff
@@ -1654,6 +1662,24 @@ def read_registration(self, authn, request, **kwargs):
...
+
+    def delete_registration(self, authn, request, **kwargs):
+        """Method to delete the client info on server side.
+
+        :param authn: Authorization HTTP header
+        :param request: Query part of the request
+        :return: Response with updated client info
+        """
+        return error_response('Unsupported operation', descr='Deletion of the registration is not supported')
```
schlenk: same as above, should be 403, according to RFC 7592 2.3

tpazderka:
OK, updated for both of you. I decoded to go the safe route and marked the `l_registration_endpoint` as deprecated, so we can remove it safely in the next release.

lwm:
🍻

PR #458, opened 2017-12-11
Optimize tox and travis

tpazderka:
---
- Using matrix only when absolutely neccessary
- Do not use pipenv for installation, it duplicates tox setup and package installation (it looks like it shortens a total runtime by 25%)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 
 sudo: false
 
+python: 3.6
+
 addons:
   apt:
     packages:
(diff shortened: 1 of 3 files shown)


lwm:
Well done!

I am not sure I understand the 'use matrix only when necessary' commit though.

Removing the `pyenv run` from the make targets goes against my initial reasoning for when using the targets when hacking - you just run `make foo` and don't need to know about pipenv. I'm open to this new approaches, but we should decide on one - some of the targets still have the `pipenv run` prefix.

tpazderka:
The python 3.5 is special as it needs to be run under pyhton3.5 the rest runs just fine with 3.6 so I used the env for the rest.

I do agree... what we can do is to separate the Makefile and the tox setup, the Makefile can still have the pipenv and the tox would have explicit commands not using Makefile?

lwm:
Ah, right! Yep, sounds good to me, nice one!

tpazderka:
OK, i will update the PR shortly :)

tpazderka:
Updated

PR #463, opened 2017-12-13
Add no-cache headers to token endpoints

### LINKED ISSUE #145: Token Endpoint should respond with nocache headers
Per [OAuth2](https://tools.ietf.org/html/rfc6749#section-5.1):

```
The authorization server MUST include the HTTP "Cache-Control"
   response header field [RFC2616] with a value of "no-store" in any
   response containing tokens, credentials, or other sensitive
   information, as well as the "Pragma" response header field [RFC2616]
   with a value of "no-cache".`
```

And [OIDC](http://openid.net/specs/openid-connect-core-1_0.html#TokenResponse):

```
All Token Responses that contain tokens, secrets, or other sensitive information MUST include the following HTTP response header fields and values:
Header Name Header Value
Cache-Control   no-store
Pragma  no-cache
```

But [oic.oic.provider:Provider.token_endpoint](https://github.com/rohe/pyoidc/blob/master/src/oic/oic/provider.py#L884) does not set these headers. While I'm using 0.7.6, it also looks like the latest master branch is the same in this regard.

In my Provider, subclass, I created a decorator for `token_endpoint` to enforce this. Others may find it useful.

``` python
# HTTP headers required with successful token responses, per OAuth2 Spec
# https://tools.ietf.org/html/rfc6749#section-5.1
OAUTH2_NOCACHE_HEADERS = (
    ('Pragma', 'no-cache'),
    ('Cache-Control', 'no-store'),
)


def nocache_if_successful_token_response(token_endpoint):
    """
    Decorate Provider.token_endpoint to ensure compliance with
    https://tools.ietf.org/html/rfc6749#section-5.1
    by adding nocache response headers if needed
    """
    def wrapped(*args, **kwargs):
        response = token_endpoint(*args, **kwargs)
        return _ensure_token_response_nocache(response) or response
    return wrapped


def _ensure_token_response_nocache(response,
                                   required_headers=OAUTH2_NOCACHE_HEADERS):
    """
    Ensure a given token response is has no cache headers that it is required
    to have per the OAuth2 Spec. This mutates the provided response. :(

    The authorization server MUST include the HTTP "Cache-Control"
    response header field [RFC2616] with a value of "no-store" in any
    response containing tokens, credentials, or other sensitive
    information, as well as the "Pragma" response header field [RFC2616]
    with a value of "no-cache".

    :param response:
    :type response: oic.utils.http_util:Response
    :rtype: oic.utils.http_util:Response
    """
    successful = ( response.status == '200 OK' )
    contains_sensitive_info = ( '"access_token"' in response.message )
    if not (successful and contains_sensitive_info):
        # Doesn't need cache headers
        return response
    existing_headers = (h for h,v in response.headers)
    for header, value in OAUTH2_NOCACHE_HEADERS:
        # Only set headers if they weren't already set
        if header not in existing_headers:
            response.headers.append((header, value))
    return response

from oic.oic.provider import Provider
class MyProvider(Provider):
    @nocache_if_successful_token_response
    def token_endpoint(self, *args, **kwargs):
        response = super(MyProvider, self).token_endpoint(*args, **kwargs)
        return response
```

I'm not in a position to make a tested PR right away, but will strive to upstream some things like this once my subclass is polished a bit.

tpazderka:
Close #145

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#436] Fixed client.read_registration
 - [#446] Fixed provider.read_registration
 - [#449] Fixed creation of error_response on client registration
+- [#145] Successful token endpoint responses have correct no-cache headers
 
 [#430]: https://github.com/OpenIDC/pyoidc/pull/430
 [#427]: https://github.com/OpenIDC/pyoidc/pull/427
@@ -33,6 +34,7 @@ The format is based on the [KeepAChangeLog] project.
 [#449]: https://github.com/OpenIDC/pyoidc/issues/449
 [#134]: https://github.com/OpenIDC/pyoidc/issues/134
 [#457]: https://github.com/OpenIDC/pyoidc/issues/457
+[#145]: https://github.com/OpenIDC/pyoidc/issues/145
 
 ## 0.12.0 [2017-09-25]
 
--- a/src/oic/extension/provider.py
+++ b/src/oic/extension/provider.py
@@ -51,6 +51,7 @@ from oic.utils.authn.client import AuthnFailure
 from oic.utils.authn.client import UnknownAuthnMethod
 from oic.utils.authn.client import get_client_id
 from oic.utils.authn.client import valid_client_info
+from oic.utils.http_util import OAUTH2_NOCACHE_HEADERS
 from oic.utils.http_util import BadRequest
 from oic.utils.http_util import Forbidden
 from oic.utils.http_util import NoContent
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#436] Fixed client.read_registration
 - [#446] Fixed provider.read_registration
 - [#449] Fixed creation of error_response on client registration
+- [#145] Succesfull token endpoint response have correct no-cache headers
```
schlenk: Successful with one 'l' only..., and should be 'responses' or 'has' (singular/plural).

tpazderka: I always get this wrong :)

Spelling fixed, I will merge it after tests pass.

PR #465, opened 2017-12-13
Refactor Message._add_value and Message.from_dict a little bit

### LINKED ISSUE #464: Decrease complexity of oic.oauth2.message.Message._add_value
Part of  #311:
`src/oic/oauth2/message.py:378:1: C901 'Message._add_value' is too complex (37) [mccabe]`

I will separate the code into separate subfunctions based on type and/or refactor it a little bit.

The function is basically untested but private so I will probably try to write tests through `from_dict`.

The max-xomplexity can then be lowered to `32`.

tpazderka:
Decreasing complexity
Close #464

---

### DIFF
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -344,90 +344,30 @@ class Message(MutableMapping):
                 continue
 
             skey = str(key)
-            try:
-                (vtyp, req, _, _deser, null_allowed) = _spec[key]
-            except KeyError:
+            if key in _spec:
+                lookup_key = key
+            else:
                 # might be a parameter with a lang tag
                 try:
                     _key, lang = skey.split("#")
                 except ValueError:
-                    try:
-                        (vtyp, _, _, _deser, null_allowed) = _spec['*']
-                        if val is None:
-                            self._dict[key] = val
-                            continue
-                    except KeyError:
-                        self._dict[key] = val
-                        continue
+                    lookup_key = '*'
                 else:
-                    try:
-                        (vtyp, req, _, _deser, null_allowed) = _spec[_key]
-                    except KeyError:
-                        try:
-                            (vtyp, _, _, _deser, null_allowed) = _spec['*']
-                            if val is None:
-                                self._dict[key] = val
-                                continue
-                        except KeyError:
-                            self._dict[key] = val
-                            continue
-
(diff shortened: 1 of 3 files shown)


tpazderka:
Touch of refactoring and some more tests.

src/oic/oauth2/message.py
```diff
@@ -470,6 +406,58 @@ def _add_value(self, skey, vtyp, key, val, _deser, null_allowed):
                 else:
                     self._dict[skey] = val
 
+    def _add_value_list(self, skey, vtype, key, val, _deser, null_allowed):
+        """Add value that is supposed to be a list.
```
lwm: Can you improve on this docstring? Not sure it gives a good overview.

lwm:
In general, nice one for taking this on! I am wondering - could we use some early `return` statements, in the new `_add_value_list` to avoid the if/elif/else nesting? Each condition could then become an if, and return early when the value is set. There would be no need for a final else, in this case, we just raise an error. Just a thought, maybe it won't reab any better ;)

src/oic/oauth2/message.py
tpazderka: Tried to improve, not sure if it is any better:)

tpazderka:
I have tried the explicit `return` statements, let me know what you think. Personally it looks the same to me so I am not against that.

lwm:
Ah! Nice, but I meant something more like:

``` diff
+        if isinstance(val, vtype):
+            if issubclass(vtype, Message):
+                self._dict[skey] = [val]
+                return
+
+            if _deser is not None:
+                try:
+                    self._dict[skey] = _deser(val, sformat="urlencoded")
+                    return
+                except Exception as exc:
+                    raise DecodeError(ERRTXT % (key, exc))
+
+            setattr(self, skey, [val])
+            return
+
+        if isinstance(val, list):
```

Was hoping it would reduce the complexity - however, writing it now, perhaps it is also a confusing read.

Up to you!

tpazderka:
I was considering that but it looked more confusing to me. The decrease in complexity is probably no longer that important since we got it below the next offender in line.

If we figure out we need to refactor it even more later on, than we can do it then.

tpazderka:
@lwm Can I get a review on this?

PR #467, opened 2017-12-22
Wrong issuer value in response when iss != baseurl

alejandro-perez:
I think this is an error. Issuer value should be taken from self.name and not self.baseurl (which might be different).

---


schlenk:
Looks sensible. Baseurl is used to derive the URLs for the endpoints mostly which do not have to match the issuer.

tpazderka:
Looks sensible to me as well.

Any objections @rohe?

rohe:
None, so I've merged.

PR #469, opened 2018-01-01
Allow endpoints to have query parts.

rohe:
---

### DIFF
--- a/src/oic/oauth2/util.py
+++ b/src/oic/oauth2/util.py
@@ -1,6 +1,7 @@
 import logging
 
 import six.moves.http_cookiejar as cookielib
+from future.backports.urllib.parse import urlsplit, urlunsplit, parse_qs
 from six import string_types
 from six.moves.http_cookiejar import http2time
 
@@ -57,8 +58,12 @@ def get_or_post(uri, method, req, content_type=DEFAULT_POST_CONTENT_TYPE,
     if method in ["GET", "DELETE"]:
         _qp = req.to_urlencoded()
         if _qp:
-            if '?' in uri:
-                path = uri + '&' + _qp
+            comp = urlsplit(uri)
+            if comp.query:
+                req.update(parse_qs(comp.query))
+                _query = req.to_urlencoded()
+                path = urlunsplit((comp.scheme, comp.netloc, comp.path,
+                                  _query, comp.fragment))
             else:
                 path = uri + '?' + _qp
         else:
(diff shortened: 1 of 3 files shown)


src/oic/oauth2/util.py
```diff
@@ -57,7 +57,10 @@ def get_or_post(uri, method, req, content_type=DEFAULT_POST_CONTENT_TYPE,
     if method in ["GET", "DELETE"]:
         _qp = req.to_urlencoded()
         if _qp:
-            path = uri + '?' + _qp
+            if '?' in uri:
```
tpazderka: I think it might be safer to use `urlsplit` and then append to the `query` part of the result and the use `urlunsplit`.

rohe: You can't append to the query part of a Splitresult but I've tried to something similar.

tpazderka: Yes, that is true. This works, but it can be simplified even more.

If you move the `urlunsplit` from the condition you can then keep the update part and in the `else` branch simpli assign the `_qs` to `_query`.

rohe: I get your point and I think I actually dealt with in the latest version of the PR.

selfissued:
Is this ready to merge and deploy or are there additional changes needed?

src/oic/oauth2/util.py
```diff
@@ -57,7 +61,14 @@ def get_or_post(uri, method, req, content_type=DEFAULT_POST_CONTENT_TYPE,
...
         if _qp:
-            path = uri + '?' + _qp
+            comp = urlsplit(uri)
+            if comp.query:
+                req.update(parse_qs(comp.query))
+                _query = req.to_urlencoded()
+                path = urlunsplit((comp.scheme, comp.netloc, comp.path,
+                                  _query, comp.fragment))
+            else:
+                path = uri + '?' + _qp
```
tpazderka: We can use the urlsplit/urlunsplit combo for this line as well. Since you are redoing this, it is a good idea to keep them the same.

src/oic/oauth2/util.py
tpazderka: Actually not, I have commented on the appropriate line.

src/oic/oauth2/util.py
```diff
@@ -55,9 +59,15 @@ def get_or_post(uri, method, req, content_type=DEFAULT_POST_CONTENT_TYPE,
     :return:
     """
     if method in ["GET", "DELETE"]:
-        _qp = req.to_urlencoded()
-        if _qp:
-            path = uri + '?' + _qp
+        if req.keys():
+            comp = urlsplit(uri)
+            if comp.query:
+                req = req.copy()
```
tpazderka: Please make this a first statement in the function to make it clear and hopefully error prone :)

PR #470, opened 2018-01-15
make message about missing registration attribute available to tester

zandbelt:
see: openid-certification/oidctest#70

this avoids falling into the Exception handler that tries to decode an error registration response error which would trigger another exception that isn't handled

Signed-off-by: Hans Zandbelt <hans.zandbelt@zmartzone.eu>

---


tpazderka:
Thank you for your contribution. Could you please add a test for this scenario?

Issue #471, opened 2018-01-22
Configuring non-default CA certificates for a KeyJar does not work as expected


schlenk:
The KeyJar constructor takes two TLS related arguments 'verify_ssl' (default=True) and 'ca_certs' (default: None). 

Trying to configure a non-default CA like this fails with a when fetching keys:

    kj = KeyJar(ca_certs=mycerts, verify_ssl=True)
    Client(ca_certs=mycerts, verify_ssl=True, keyjar=kj, ...)

It also fails to just use the defaults and just configure the client without an explicit keyjar:

    Client(ca_certs=mycerts, verify_ssl=True, ...)

The ca_certs parameter to KeyJar is stored, but never used. The KeyBundle Class only ever sees the verify_ssl parameter that is passed down to requests.

The default KeyJar setup in PBase.__init__ is also broken, so you need to set BOTH verify_ssl and ca_certs to a path to make it work properly. PBase gets it right for its own http_request code though, but fails to setup the default KeyJar correctly.

Looks like this in a slightly redacted traceback when doing an AccessTokenRequest with a RS256 signed response:

    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oic/__init__.py", line 653, in do_access_token_request
    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oauth2/__init__.py", line 761, in do_access_token_request
    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oauth2/__init__.py", line 684, in request_and_return
    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oauth2/__init__.py", line 638, in parse_request_response
    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oauth2/__init__.py", line 565, in parse_response
        verf = resp.verify(**kwargs)
    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oic/message.py", line 302, in verify
        idt = IdToken().from_jwt(str(self["id_token"]), **args)
    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oauth2/message.py", line 687, in from_jwt
    File "..../python2.7/site-packages/oic-0.11.0.0-py2.7.egg/oic/oauth2/message.py", line 599, in get_verify_keys
        self._add_key(keyjar, jso[ent], key, _key_type, _kid, nki)
[...]

PR #472, opened 2018-01-23
Remove ca_certs option and replace it by verify_ssl

schlenk:
Remove the ca_certs option that was not handled consistently.
Now our verify_ssl option has the same semantics as the requests verify option.

Also added a few words about TLS in the documentation.

This fixes https://github.com/OpenIDC/pyoidc/issues/471


tpazderka:
LGTM

Issue #352, opened 2017-05-19
oic.util.keyio.proper_path produces bogus results on Windows


schlenk:
The `proper_path` method in the keyio module produces broken pathes on windows, as it does not handle drive letters in any way.

This leads to a test failure on windows, when running `TestProvider.test_provider_key_setup`, as the `os.makedirs` gets some path looking like `./c:' and doesn't like that.

I'm not sure what the code should really do, as the desired usage of `proper_path` isn't really clear to me, it looks a bit like it is mostly wishful thinking. It seems to try to produce some kind of relativ path (which might be impossible for windows sometimes, e.g. if files reside on different drive letters).

The usage seems a little odd too, especially in `key_export()`. It doesn't make much sense to concatenate the exported URL path with some local path and expect the result to be sane. Those two pathes might have no relation at all besides some mapping in some web server or reverse proxy configuration.

It would help to document the expectations/path vs URL layout ideas behind the keyio code a bit.

lwm:
> I'm not sure what the code should really do, as the desired usage of proper_path isn't really clear to me, it looks a bit like it is mostly wishful thinking. It seems to try to produce some kind of relativ path (which might be impossible for windows sometimes, e.g. if files reside on different drive letters).

> The usage seems a little odd too, especially in key_export(). It doesn't make much sense to concatenate the exported URL path with some local path and expect the result to be sane. Those two pathes might have no relation at all besides some mapping in some web server or reverse proxy configuration.

@rohe, this is certainly a job for you!

rohe:
I'm been using Unix or Unix like OS's since late 80's so I have no recent experience of Windows or similar systems. There have also not been any pressure from my immediate surrounding to deal with anything but Unix like system. Due to this I'm sure there are a number of things that doesn't work on Windows. If you want to fix this please go ahead.

Regarding the other issue; yes I have assumed that there is a simple map between a URL and a local path. Again this is the environment I've worked and is working within.
I do belive there has to be some function you can use to go from one to the other, albeit it might be more complex then I believed it would have to be.

schlenk:
ok. Yes, just one OS tradition makes things so much simpler. 

I'll have a look and see if i can figure out the assumptions. Should not be too many (i have it running on Windows, so it works in principle, just some tests blow up currently, and to fix those i need the assumptions.)

tpazderka:
@schlenk Any luck on this? I personally think that we should rethink the `keyio` as a whole since it assume so many things.

PR #473, opened 2018-01-24
Fix broken test on windows

schlenk:
Fix the test_provider_key_setup test.

The code under test makes bad assumptions about the filesystem layout.

Basically key_setup/key_export only work for a very specific setup where the export dir is inside the web servers root directory.

The test ignored this assumption and just tried to stuff any random tmpdir path there, which does not work.

The fix just adjusts the tempdir path to a local name, that works.

This should kind of fix https://github.com/OpenIDC/pyoidc/issues/352.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#445] Fixed get_client_id
 - [#421] Fixed handling of unicode in sanitize function
 - [#145] Successful token endpoint responses have correct no-cache headers
+- [#352] Fixed broken windows test for ``test_provider_key_setup``. 
 
 [#430]: https://github.com/OpenIDC/pyoidc/pull/430
 [#427]: https://github.com/OpenIDC/pyoidc/pull/427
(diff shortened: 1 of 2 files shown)


schlenk:
@tpazderka Maybe you could give appveyor a kick, so it tests this? It should fix the last permanently failing test on windows.

tpazderka:
@schlenk I have tried to do it but it still refuses to build automatically from master and PR... not sure why and why it worked previously.

tpazderka:
Can you try a rebase on current master?

tpazderka:
Still not triggering. I have asked the AppVeyor staff for support. I will get back to this after we can trigger the builds.

schlenk:
Thats fine, its mostly cosmetics to get a clean test run, no real features involved. 

The keyio API for key_export/key_setup is just broken, so the test simply uses a 'working' path.

tpazderka:
OK, the AppVeyor builds are up and running. They will apply only for a new PR, so this one won't get it.

I have no issue with this change, but I would like to think about getting rid of this API, since it doesn't make too much sense...

PR #479, opened 2018-02-02
Base implementation of ClientDatabase classes

### LINKED ISSUE #478: Add Baseclass for ClientDatabase
There is no BaseClass for implementation of custom client databases. We should create one that defines the basic expected API.

The class should probably reside in `oic.utils.cdb` and the basic API can be based on `oic.utils.client_management.CDB` object which can then inherit from the new class.

The basic expected API should be along these lines:
```
class BaseClientDatabase(object)
    def __init__(self)
    def get(self, item, default=None)
    def __getitem__(self, item)
    def __setitem__(self, item, value)
    def __delitem__(self, key)
    def __contains__(self, item)
    def keys(self)
    def items(self)
   def __len__(self)
```
The `__init__` methods in providers can then warn if the `cdb` argument doesn't inherit from the new `BaseClientDatabase`.

tpazderka:
Close #478

---

### DIFF
--- a/src/oic/utils/clientdb.py
+++ b/src/oic/utils/clientdb.py
@@ -1,12 +1,16 @@
 """Client managament databases."""
+from abc import ABCMeta
+from abc import abstractmethod
+
 import requests
+from six import with_metaclass
 from six.moves.urllib.parse import quote
 from six.moves.urllib.parse import urljoin
 
 from oic.oauth2.exception import NoClientInfoReceivedError
 
 
-class BaseClientDatabase(object):
+class BaseClientDatabase(with_metaclass(ABCMeta)):
     """
     Base implementation for Client management database.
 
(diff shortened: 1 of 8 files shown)


tpazderka:
I am not sure about the implementation of `items` and `keys` in `MDQClient` as I cannot find any working example of `MDQ` database.
@rohe Do you have any examples of this so I can test the implementation?

src/oic/utils/clientdb.py
```diff
@@ -1,22 +1,134 @@
...
+"""Client managament databases."""
 import requests
 from six.moves.urllib.parse import quote
+from six.moves.urllib.parse import urljoin
 
 from oic.oauth2.exception import NoClientInfoReceivedError
 
 
-class MDQClient(object):
+class BaseClientDatabase(object):
```
schlenk: Would it make any sense to use abc.ABCMeta here to declare the baseclass and get rid of the NotImplementedErrors in exchange for @abstractmethod decorators?

tpazderka: I never used this approach, so I have no experience with it. I will look into this on Monday.

tpazderka:
OK, redone as `ABCMeta`, it actually looks better :)

PR #482, opened 2018-02-05
Validate expiration of Token at UserInfo endpoint

### LINKED ISSUE #424: userinfo_endpoint return data if access_token is expired

tpazderka:
Close #424

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#145] Successful token endpoint responses have correct no-cache headers
 - [#352] Fixed broken windows test for ``test_provider_key_setup``. 
 - [#475] ``get_verify_key`` returns inactive ``sig`` keys for verification
+- [#429] An expired token is not possible to use.
 
 [#430]: https://github.com/OpenIDC/pyoidc/pull/430
 [#427]: https://github.com/OpenIDC/pyoidc/pull/427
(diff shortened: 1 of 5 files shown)


schlenk:
Seems there is a test failure in the new test, build is flagged as broken on all Python versions.

>       assert message['error'] == 'invalid_token'
E       KeyError: 'error'

tpazderka:
That's weird as it was passing before. I will look into this today.

schlenk:
I had a short look, seems the `is_expired()` API of sdb.Token is kind of broken by design.

The code calls `is_expired()` on the factory (Token is badly named, its actually a TokenFactory...), not on the token, and does not pass in the token, so i cannot imagine how it could work with this API.

As the DefaultToken does not pass in the expiry date / issued at / lifetime into the token in __call__() this information is lost, unless it is stored out-of-band inside the SDB. Which is probably wrong and DefaultToken should encode the expiry time/time of issuance in the token. So the is_expired() would need to be on the derived Token classes.

The test passes if one increases the timespan to 1yr (year 2019) btw., which is extra weird.

tpazderka:
Ah, I see. I noticed that the API is broken and tried to fix it with the `issued_at` but apparently failed :)

Weird thing is that tests passed on the branch and also after the merge but started to fail as of yesterday...

I guess we can encode the issue datetime to the token itself and then it should work.

schlenk:
I have a fix nearly ready, will open a PR for it soon.

tpazderka:
OK, I would get to this today but if you are already done, I can review it.

Issue #486, opened 2018-02-09
SystemRandom is not imported correctly


shelman:
In `src/oic/__init__.py`, `random.SystemRandom` is imported via the following code:
```
try:
    import random.SystemRandom as rnd
except ImportError:
    import random as rnd
```

In both python2 and python3 this yields an import error and falls back to the `random` package.

I believe the fix is to change the import to `from random import SystemRandom as rnd`. The import works locally for me but I am having some trouble with the unit tests and wanted to check in before opening a pull request.

schlenk:
Thats bad. Yes, PR welcome. That shouldn't break any tests.

PR #487, opened 2018-02-12
Fix broken import of SystemRandom

schlenk:
This fixes https://github.com/OpenIDC/pyoidc/issues/486 .

This is a pretty bad security issue, as the secrets used to initialize most crypto is created by `rndstr`, so one can probably brute force things with some small effort (e.g. using untwister and some known values).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,12 +29,8 @@ pyoidc*
 foo.*
 
 # Windows testing stuff
-.pytest_cache/
 .coverage
 .cache/
-appdata/
-users/
-
 
 # Dynamically created doc folders
 doc/_build
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,14 @@ secret/
 pyoidc*
 foo.*
 
+# Windows testing stuff
+.pytest_cache/
+.coverage
+.cache/
+appdata/
+users/
+
+
 # Dynamically created doc folders
 doc/_build
 !tests/data/keys
--- a/src/oic/__init__.py
+++ b/src/oic/__init__.py
@@ -3,9 +3,22 @@ import string
 
 # Since SystemRandom is not available on all systems
 try:
-    import random.SystemRandom as rnd
+    # Python 3.6+, designed for this usecase
+    from secrets import choice
 except ImportError:
-    import random as rnd
+    import random
+    try:
+        # Python 2.4+ if available on the platform
+        _sysrand = random.SystemRandom()
+        choice = _sysrand.choice
+    except AttributeError:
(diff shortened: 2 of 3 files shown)


src/oic/__init__.py
```diff
@@ -3,9 +3,22 @@
...
+    # Python 3.6+, designed for this usecase
+    from secrets import choice
 except ImportError:
-    import random as rnd
+    import random
+    try:
+        # Python 2.4+ if available on the platform
+        _sysrand = random.SystemRandom()
+        choice = _sysrand.choice
+    except AttributeError:
```
tpazderka: I don't think that we officially support `python < 2.7` so I think this `except` branch is not really necessary.

src/oic/__init__.py
```diff
@@ -3,9 +3,22 @@
 
 # Since SystemRandom is not available on all systems
 try:
-    import random.SystemRandom as rnd
+    # Python 3.6+, designed for this usecase
+    from secrets import choice
```
tpazderka: I believe that even this should be initialized with a call to `SystemRandom`.

.gitignore
```diff
@@ -28,6 +28,14 @@ secret/
 pyoidc*
 foo.*
 
+# Windows testing stuff
```
tpazderka: Wow this file is a mess... This should really contain only files that are regularly generated by python itself. All the other files generated by IDE or OS should be a part of user `.gitignore`

I guess that `.cache` and `.coverage` can remain here but the rest should go to your personal `.gitignore`.

src/oic/__init__.py
schlenk: True, but even on Python 2.7 the docs state that some platforms lack SystemRandom().

Not that it would be a good idea to deploy this code on any such platform...

src/oic/__init__.py
schlenk: Not needed. The secrets module does that under the hood.

See 
https://github.com/python/cpython/blob/6ea20fc719dcaf102d2cefa1786f0421514f7a58/Lib/secrets.py#L22

.gitignore
schlenk: Agreed. Will do.

src/oic/__init__.py
tpazderka: OK.

src/oic/__init__.py
tpazderka: OK

tpazderka:
Please rebase on current master and we can proceed with this one.

PR #488, opened 2018-02-12
Better fix for token expiry check

schlenk:
Encoded the issued at timestamp in the DefaultToken, so we can check it later.

The specific factories are responsible for checking the tokens they issued.

This fixes the broken tests introduced with PR #482.

Old DefaultToken issued tokens will need to be reissued, JWTTokens stay valid with this change. But due to the problems with randomness it is a good idea to re-issue all tokens anyway.

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -1221,8 +1221,9 @@ class Provider(AProvider):
             logger.error('Wrong token type: {}'.format(typ))
             raise FailedAuthentication("Wrong type of token")
 
-        if _sdb.access_token.is_expired():
-            return error(error='invalid_token', descr='Token is expired', status_code=401)
+        if _sdb.access_token.is_expired(token):
+            return error(error='invalid_token', descr='Token is expired',
+                         status_code=401)
 
         if _sdb.is_revoked(key):
             return error(error="invalid_token", descr="Token is revoked",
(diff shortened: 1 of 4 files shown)


src/oic/oic/provider.py
```diff
@@ -1221,8 +1221,9 @@ def _do_user_info(self, token, **kwargs):
             logger.error('Wrong token type: {}'.format(typ))
             raise FailedAuthentication("Wrong type of token")
 
-        if _sdb.access_token.is_expired():
-            return error(error='invalid_token', descr='Token is expired', status_code=401)
+        if _sdb.access_token.is_expired(token):
+            return error(error='invalid_token', descr='Token is expired',
```
tpazderka: Why are you changing this? Not that it matters but we have a line length set to 120 chars and this is certainly below that limit :)

schlenk: Right, just looked two lines down, where a similar check was wrapped.

tpazderka: Yes, I noticed that as well. I think it is a difference between old code (which had a limit to 80) and new/changed code...

schlenk:
Ah, those extra files should have been ignored.

src/oic/oic/provider.py
schlenk: Can revert if you care. Will look for the 120 thing in the future, my default is usually 80.

tpazderka: No, I don't care enough for a revert.

tpazderka:
I guess that in light of this, we should really cleanup the sdb module...

PR #493, opened 2018-02-16
grant_types should follow response_types in a client registration req…

rohe:
grant_types should follow response_types in a client registration request

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on the [KeepAChangeLog] project.
 ## 0.13.0 [Unreleased]
 
 ### Added
+- [#493] grant_types specification should follow the response_types sprecification in a client registration request.
 - [#469] Allow endpoints to have query parts
 - [#443] Ability to specify additional supported claims for oic.Provider
 - [#134] Added method kwarg to registration_endpoint that enables the client to read/modify registration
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on the [KeepAChangeLog] project.
 ## 0.13.0 [Unreleased]
 
 ### Added
+- [#493] grant_types specification should follow the response_types sprecification in a client registration request.
```
schlenk: specification, not sprecification

src/oic/oic/__init__.py
```diff
@@ -257,6 +257,32 @@ def add_token(self, resp):
...
+    'code': ['authorization_code'],
+    'id_token': ['implicit'],
+    'id_token token': ['implicit'],
+    'code id_token': ['authorization_code', 'implicit'],
+    'code token': ['authorization_code', 'implicit'],
+    'code id_token token': ['authorization_code', 'implicit']
+}
+
+
+def response_types_to_grant_types(response_types):
```
tpazderka: This should be covered by tests.

src/oic/oic/__init__.py
```diff
@@ -1319,6 +1345,12 @@ def create_registration_request(self, **kwargs):
         except KeyError:
             pass
 
+        try:
```
tpazderka: I think it is better to change `try .. except`  to a condition `if response_types in req`.

src/oic/oic/__init__.py
schlenk: Especially the part that response_types may be in random order (so no one removes the 'sort' from the code by accident).

CHANGELOG.md
rohe: Right

src/oic/oic/__init__.py
rohe: I'm OK with either or so we'll go for your proposal.

src/oic/oic/__init__.py
tpazderka: One last thing then this is good to go.

The test with an unknown `response_type ` should also be present.

rohe: Added

PR #506, opened 2018-03-13
Fix UserInfo endpoint for deleted clients

### LINKED ISSUE #503: Crash at Introspection and UserInfo endpoints with token of deleted client
Use case:
1) Get access token for client
2) Delete client
3) Try to verify token (introspection endpoint) or get user info

tpazderka:
Close #503

---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -1202,7 +1202,9 @@ class Provider(AProvider):
         info = self.schema(**self._collect_user_info(session))
 
         # Should I return a JSON or a JWT ?
-        _cinfo = self.cdb[session["client_id"]]
+        _cinfo = self.cdb.get(session["client_id"])
+        if _cinfo is None:
+            return error_response("unauthorized_client", descr="Unknown client")
         try:
             if "userinfo_signed_response_alg" in _cinfo:
                 # Will also encrypt if defined in cinfo
--- a/tests/test_oic_provider.py
+++ b/tests/test_oic_provider.py
@@ -829,6 +829,32 @@ class TestProvider(object):
         ident = OpenIDSchema().deserialize(resp.message, "json")
         assert _eq(ident.keys(), ['nickname', 'sub', 'name', 'email'])
 
+    def test_userinfo_endpoint_missing_client(self):
+        self.provider.cdb["unknownclient"] = {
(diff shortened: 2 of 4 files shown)


src/oic/oic/provider.py
```diff
@@ -1202,7 +1202,9 @@ def _do_user_info(self, token, **kwargs):
         info = self.schema(**self._collect_user_info(session))
 
         # Should I return a JSON or a JWT ?
-        _cinfo = self.cdb[session["client_id"]]
+        _cinfo = self.cdb.get(session["client_id"])
+        if _cinfo is None:
+            return error_response('unauthorized_client', descr='Unknown client')
```
a-wakeel: This could be handled inside the **except** block. Isn't?

a-wakeel: This could be handled inside the **except** block. Isn't?

tests/test_oic_provider.py
```diff
@@ -829,6 +829,32 @@ def test_userinfo_endpoint_authn(self):
         ident = OpenIDSchema().deserialize(resp.message, "json")
         assert _eq(ident.keys(), ['nickname', 'sub', 'name', 'email'])
 
+    def test_userinfo_endpoint_missing_client(self):
+        self.provider.cdb["unknownclient"] = {
+            "client_secret": "unknownclient",
+            "redirect_uris": [("http://localhost:8087/authz", None)],
+            "post_logout_redirect_uris": [("https://example.com/post_logout", None)],
+            "client_salt": "salted",
+            'response_types': ['code', 'token', 'code id_token', 'none', 'code token', 'id_token']
```
a-wakeel: Is there any specific reason for using **single** quotes here, as **double** quotes are used in other key pairs?

a-wakeel: Same applies for other usages

src/oic/oic/provider.py
tpazderka: Yes, but not all implementations of ClientDatabase raise `KeyError`, so I went with safer approach.

tests/test_oic_provider.py
tpazderka: No specific. We do have an inconsistent quoting style :(

I will match the quotes inside a test to match.

PR #510, opened 2018-03-21
Remove duplicate mapping in Client_db

### LINKED ISSUE #499: What kind of structure "cdb" should be ?
https://github.com/OpenIDC/pyoidc/blob/3f21791df99aad8764b3ce796ba5689241ca86c4/src/oic/utils/authn/client.py#L490-L501

We got get_client_id utility where inhomogeneous structure of cdb:
`{
'client_id': client info',
'token': 'client_id'
}`

It's normal or this is a bug ?

### LINKED ISSUE #507: Drop duplicate structure of CDB
CDB now needs to keep two mappings for clients. One for `client_id` and one for `_rat`.

The access by `_rat` is used only in `oic.oic.Provider.read_registration` where it can be easily removed by switching the order of checks.

The client info would be retrieved using `client_id` in request and the correct `_rat`  will be checked.

Brought up in #499 as well.

tpazderka:
Close #507
Close #499

---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -23,6 +23,7 @@ import six
 from jwkest import b64d
 from jwkest import jwe
 from jwkest import jws
+from jwkest import safe_str_cmp
 from jwkest.jwe import JWE
 from jwkest.jwe import JWEException
 from jwkest.jwe import NotSupportedAlgorithm
@@ -1533,8 +1534,6 @@ class Provider(AProvider):
             "client_salt": rndstr(8)
         }
 
-        self.cdb[_rat] = client_id
-
         _cinfo = self.do_client_registration(request, client_id,
                                              ignore=["redirect_uris",
                                                      "policy_uri", "logo_uri",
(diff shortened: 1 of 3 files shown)


src/oic/oic/provider.py
```diff
@@ -1594,14 +1592,11 @@ def read_registration(self, authn, request, **kwargs):
...
-        except KeyError:
-            return Unauthorized()
-
-        # extra check
+        # Get client_id from request
         _info = parse_qs(request)
-        if not _info["client_id"][0] == client_id:
+        client_id = _info.get("client_id", [None])[0]
+
+        if self.cdb.get(client_id) is None or self.cdb.get(client_id)['registration_access_token'] != token:
```
schlenk: The access token comparision should use a constant time compare function, to avoid a timing attack.
It should also ALWAYS check the access token against a dummy value, if the client_id does not exist.

    from jwkest import safe_str_cmp
    cdb_entry = self.cdb.get(client_id)
    reg_token = cdb_entry['registration_acccess_token'] if cdb_entry else ''
    token_broken = not safe_str_cmp(reg_token, token)
    if cdb_entry is None or token_broken:
        ...

Otherwise someone could do a timing attack to guess the registration_access_token and get access to the client_secret. This wasn't possible with the old style, as that did a hash on the token, which basically is constant time and looked it up in the cdb.

tpazderka:
OK, I have added constant time checking.

The implementation of `get_client_id` is probably wrong. It should really rely on `client_id` in request (it is a required param). I think we can drop the whole `elif == "bearer"` branch... Any objections?

schlenk:
I am not sure if the bearer part in get_client_id is actually used, its a bit hard to find out, but i doubt it. I don't see a case where you do not have the client_id available too.

The rest looks good though.

schlenk:
I think the get_client_id fix should be a different issue.

tpazderka:
OK, I will create an issue for that.

Btw, I have added a missing test...

PR #518, opened 2018-04-09
fix keyio

ermakoves:
---

### DIFF
--- a/src/oic/utils/keyio.py
+++ b/src/oic/utils/keyio.py
@@ -546,12 +546,15 @@ class KeyJar(object):
                     if kid and key.kid == kid:
                         lst = [key]
                         break
-                    if not key.use or use == key.use:
-                        lst.append(key)
-                        break
+
                     # Verification can be performed by both `sig` and `ver` keys
                     if key_use == 'ver' and key.use in ('sig', 'ver'):
                         lst.append(key)
+                        continue
+
+                    if not key.use or use == key.use:
+                        lst.append(key)
+                        break
 
         # if elliptic curve have to check I have a key of the right curve
         if key_type == "EC" and "alg" in kwargs:
--- a/src/oic/utils/keyio.py
+++ b/src/oic/utils/keyio.py
@@ -546,15 +546,12 @@ class KeyJar(object):
                     if kid and key.kid == kid:
                         lst = [key]
                         break
-
+                    if not key.use or use == key.use:
+                        lst.append(key)
+                        continue
                     # Verification can be performed by both `sig` and `ver` keys
                     if key_use == 'ver' and key.use in ('sig', 'ver'):
                         lst.append(key)
-                        continue
-
-                    if not key.use or use == key.use:
-                        lst.append(key)
-                        break
 
         # if elliptic curve have to check I have a key of the right curve
         if key_type == "EC" and "alg" in kwargs:
--- a/tests/test_keyio.py
+++ b/tests/test_keyio.py
@@ -352,13 +352,13 @@ class TestKeyJar(object):
         assert _key
         assert _key.kid == "abc"
 
-        def test_get_inactive_ver(self):
-            ks = KeyJar()
-            ks['http://example.com'] = KeyBundle([{"kty": "oct", "key": "a1b2c3d4", "use": "sig"},
-                                                  {"kty": "oct", "key": "a1b2c3d4", "use": "ver"}])
-            ks['http://example.com'][0]._keys[1].inactive_since = 1
-            key = ks.get_verify_key(owner='http://example.com')
-            assert len(key) == 2
+    def test_get_inactive_ver(self):
+        ks = KeyJar()
+        ks['http://example.com'] = KeyBundle([{"kty": "oct", "key": "a1b2c3d4", "use": "sig"},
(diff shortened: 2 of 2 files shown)


tests/test_keyio.py
```diff
@@ -352,13 +352,13 @@ def test_get_by_kid(self):
...
         assert _key.kid == "abc"
 
-    def test_get_inactive_ver(self):
-        ks = KeyJar()
-        ks['http://example.com'] = KeyBundle([{"kty": "oct", "key": "a1b2c3d4", "use": "ver"}])
-        ks['http://example.com'][0]._keys[0].inactive_since = 1
-        key = ks.get_verify_key(owner='http://example.com')
-
-        assert len(key) == 1
+        def test_get_inactive_ver(self):
```
tpazderka: This is an extra indent.

tests/test_keyio.py
```diff
@@ -303,8 +303,8 @@ def test_keyjar_group_keys(self):
             keybundle_from_local_file(RSAKEY, "rsa", ["ver", "sig"]))
 
         verified_keys = ks.verify_keys("http://www.example.org")
-        assert len(verified_keys) == 4
-        assert len([k for k in verified_keys if k.kty == "oct"]) == 2
+        assert len(verified_keys) == 6
+        assert len([k for k in verified_keys if k.kty == "oct"]) == 4
```
tpazderka: This seems to be wrong. The total number of keys in this `KeyJar` with `oct` type is 2. So this result contains some keys multiple times.

ermakoves: Maybe. So there is the problem of method x_keys:
https://github.com/OpenIDC/pyoidc/blob/d0d502b7f479a3d721817e96f16278a4f1b9c477/src/oic/utils/keyio.py#L608-L613
line 611 - 4 keys, next will be 6

ermakoves: In total you have received keys for all owners

tpazderka: I am not sure what the reasoning for this is. Maybe a symmetric key lookup?
@rohe Can you please look into this?

tpazderka: @rohe, can you please have a look? I think that this might be correct since we want to include our keys as well but I am not sure.

rohe: When it comes to symmetric keys we definitely wants to include our own keys.
Looking at x_keys now I see at least one problem and that is as you say the problem of getting multiple copies of the same key. Since jwkest.jwk.Key instance has an __eq__ method we should be able to weed out those.
One reason why we get multiple copies is because if a key is marked to be for both signing and encrypting it is handled as 2 keys. And in the test the same key is marked to be both for signing and verification which is overkill since it's sufficient for a key to be mark for 'ver' to be used for both. Some cleaning up is in order.

tpazderka: @ermakoves Can you please accommodate for the comment by @rohe so we can close this one?

ermakoves: @tpazderka I'm not quite sure what I need to do. Share keys in the test for signed and encrypted? If so, then I do not think that this will clarify, and besides, why can not the key be able to do everything?

tpazderka: @ermakoves Well I had another look at your tests (`test_get_inactive_ver`) and they pass for me even with the existing code. Could you please provide us with a test that is failing on the current master and shoulc pass according to you?

ermakoves: run test on your master
>tests\test_keyio.py:354 (TestKeyJar.test_get_inactive_ver)
>self = <tests.test_keyio.TestKeyJar object at 0x051F3310>
>
>   def test_get_inactive_ver(self):
>        ks = KeyJar() 
>        ks['http://example.com'] = KeyBundle([{"kty": "oct", "key": "a1b2c3d4", "use": "sig"},
>                                              {"kty": "oct", "key": "a1b2c3d4", "use": "ver"}])    
>        ks['http://example.com'][0]._keys[1].inactive_since = 1   
>        key = ks.get_verify_key(owner='http://example.com')
>       assert len(key) == 2
>E       assert 1 == 2
>E        +  where 1 = len([<jwkest.jwk.SYMKey object at 0x051F3B10>])
>
>test_keyio.py:364: AssertionError'''

ermakoves: 2018-05-14_18-07_d-git-forks-pyoidc_origin-pyo

ermakoves: The keys are not duplicates in another test just returns all valid keys for all owners

tpazderka: Yes, this is not a case of tests only, but a general principle. We can deal with it in different issue later on.

src/oic/utils/keyio.py
```diff
@@ -546,12 +546,15 @@ def get(self, key_use, key_type="", issuer="", kid=None, **kwargs):
...
-                        break
+
                     # Verification can be performed by both `sig` and `ver` keys
                     if key_use == 'ver' and key.use in ('sig', 'ver'):
                         lst.append(key)
+                        continue
+
+                    if not key.use or use == key.use:
+                        lst.append(key)
+                        break
```
tpazderka: This is not needed as if you would have a key without use, you won't be able to gather any other keys.

tpazderka:
Also, please rebase on current master, then we can merge this.

PR #533, opened 2018-04-29
Quote

rohe:
---

There's a problem with the way client_secret_basic client authentication is done. 
According to https://tools.ietf.org/html/rfc6749#section-2.3.1 both Client id and client secret must be url encoded before concatenated into the authorisation token.


rohe:
What do you want me to do ? Start all over with a pristine copy ?

schlenk:
I think you just need to rebase to master. The mixed up change seems to be the #493 fix, which is already merged to master.

There seem to be some test failures with the new tests though.

PR #534, opened 2018-05-03
According to https://tools.ietf.org/html/rfc6749#section-2.3.1 client…

rohe:
… identifier and secret MUST be x-www-form-urlencoded when doing client_secret_basic client authentication.

---

### DIFF
--- a/src/oic/utils/authn/client.py
+++ b/src/oic/utils/authn/client.py
@@ -110,8 +110,7 @@ class ClientSecretBasic(ClientAuthnMethod):
             http_args["headers"] = {}
 
         credentials = "{}:{}".format(quote_plus(user), quote_plus(passwd))
-        authz = base64.urlsafe_b64encode(credentials.encode("utf-8")).decode(
-            "utf-8")
+        authz = base64.b64encode(credentials.encode("utf-8")).decode("utf-8")
         http_args["headers"]["Authorization"] = "Basic {}".format(authz)
 
         try:
--- a/src/oic/utils/authn/client.py
+++ b/src/oic/utils/authn/client.py
@@ -1,8 +1,9 @@
+from future.backports.urllib.parse import quote_plus
+
 import base64
 import logging
 
 import six
-from future.backports.urllib.parse import quote_plus
 from jwkest import Invalid
 from jwkest import MissingKey
 from jwkest import as_bytes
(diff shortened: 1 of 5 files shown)


src/oic/utils/authn/client.py
```diff
@@ -108,7 +109,7 @@ def construct(self, cis, request_args=None, http_args=None, **kwargs):
         if "headers" not in http_args:
             http_args["headers"] = {}
 
-        credentials = "{}:{}".format(user, passwd)
+        credentials = "{}:{}".format(quote_plus(user), quote_plus(passwd))
         authz = base64.urlsafe_b64encode(credentials.encode("utf-8")).decode(
```
panva: To think of it, shouldn't content be just `b64encode`d instead of `urlsafe_b64encode`d?

tests/test_client.py
```diff
@@ -52,10 +53,9 @@ def test_construct(self, client):
 
         csb = ClientSecretBasic(client)
         http_args = csb.construct(cis)
-
+        cred = '{}:{}'.format(quote_plus("A"), quote_plus("boarding pass"))
         assert http_args == {"headers": {"Authorization": "Basic {}".format(
-            base64.urlsafe_b64encode("A:boarding pass".encode("utf-8")).decode(
-                "utf-8"))}}
+            base64.urlsafe_b64encode(cred.encode("utf-8")).decode("utf-8"))}}
```
panva: same, shouldn't basic auth use just `b64encode`?

tests/test_authn_user.py
```diff
@@ -15,3 +19,22 @@ def test_symkeyauthn_improperly_configured(provider):
...
+        'Diana': 'Piano player',
+        'NonAscii': '€&+%#@äää'
+    }
+    ba = BasicAuthn(None, pwd=pwd_database)
+
+    for user, passwd in pwd_database.items():
+        credentials = "{}:{}".format(quote_plus(user),
+                                     quote_plus(passwd))
+
+        authz = base64.urlsafe_b64encode(credentials.encode("utf-8")).decode(
```
panva: same, shouldn't basic auth use just `b64encode`?

schlenk:
Changelog entry is still missing, looks good otherwise.

PR #535, opened 2018-05-05
interpret passed in grant_types when assembling the registration request

zandbelt:
completes/fixes 23efe6dd0638c23e0db90e11d86634c560a92c55 

Signed-off-by: Hans Zandbelt <hans.zandbelt@zmartzone.eu>

---

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -268,8 +268,11 @@ rt2gt = {
 }
 
 
-def response_types_to_grant_types(response_types):
+def response_types_to_grant_types(response_types, kwargs):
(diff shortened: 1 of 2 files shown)


src/oic/oic/__init__.py
```diff
@@ -268,9 +268,12 @@ def add_token(self, resp):
 }
 
 
-def response_types_to_grant_types(response_types):
+def response_types_to_grant_types(response_types, kwargs):
```
schlenk: kwargs is unusual as a single argument, python style would be **kwargs.

probably rename it to extra_grant_types

PR #541, opened 2018-05-24
Avoid TypeError in Python 3.x from integer status codes

claweyenuk:
---

Ran into this small bug running in Python 3 for redirect responses.


claweyenuk:
Thanks for the quick review!  I pushed a fix

Issue #212, opened 2016-07-08
Try to login across simple oidc_exampla and get error on login 


demchenkoa:
When enter name and pass "diana" "krall", redirect me to https://localhost:8092/verify and print traceback 
WSGI response header value u'https://localhost:8092/authorization?acr_values=PASSWORD&redirect_uri=http%3A%2F%2Flocalhost%3A8666%2F919D3F697FDAAF138124B83E09ECB0B7&response_type=code&client_id=ybnRmeIpZjrm&upm_answer=true&scope=openid+profile+email+address+phone' is not of type str.
['Traceback (most recent call last):\n', '  File "/home/ademchenko/pyoidc-master/oidc_example/op2/server.py", line 361, in application\n    return callback(environ, start_response, logger)\n', '  File "/home/ademchenko/.virtualenvs/pyoidc/local/lib/python2.7/site-packages/oic-0.8.5.1-py2.7.egg/oic/utils/authn/authn_context.py", line 217, in auth_verify\n    return response(environ, start_response)\n', '  File "/home/ademchenko/.virtualenvs/pyoidc/local/lib/python2.7/site-packages/oic-0.8.5.1-py2.7.egg/oic/utils/http_util.py", line 123, in __call__\n    start_response(self.status, self.headers)\n', '  File "/home/ademchenko/.virtualenvs/pyoidc/local/lib/python2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 2297, in start_response\n    "WSGI response header value %r is not of type str." % v)\n', "TypeError: WSGI response header value u'https://localhost:8092/authorization?acr_values=PASSWORD&redirect_uri=http%3A%2F%2Flocalhost%3A8666%2F919D3F697FDAAF138124B83E09ECB0B7&response_type=code&client_id=ybnRmeIpZjrm&upm_answer=true&scope=openid+profile+email+address+phone' is not of type str.\n"]
AssertionError('WSGI start_response called a second time with no exc_info.',)
Traceback (most recent call last):
  File "/home/ademchenko/.virtualenvs/pyoidc/local/lib/python2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 1360, in communicate
    req.respond()
  File "/home/ademchenko/.virtualenvs/pyoidc/local/lib/python2.7/site-packages/cherrypy/wsgiserver/wsgiserver2.py", line 875, in respond
    self.server.gateway(self).respond()
[...]

PR #542, opened 2018-05-24
Bugfix/working simple examples

claweyenuk:
---

I don't think this is mergeable, but I don't have the time to clean this up at the moment, particularly the keyjar code.  But I thought someone might find this useful since there are issues like #212 and it is mentioned in the main README.

The steps in the README in here took some time to figure out as well, so hopefully that can help some people.

Note that I tested this using Python 3.  I did run this under Python 2 and it required a few very bad hacks to get working in the wsgi lib that I think are because start response may have been called twice.  I did not dig into that

### DIFF
--- a/oidc_example/simple_op/requirements.txt
+++ b/oidc_example/simple_op/requirements.txt
@@ -1,6 +1,5 @@
 cherrypy==3.2.4
 pyaml==15.03.1
 Jinja2==2.7.3
-oic==0.7.6
 yubico-client==1.9.1
 pyjwkest==1.0.1
--- a/oidc_example/simple_rp/requirements.txt
+++ b/oidc_example/simple_rp/requirements.txt
@@ -1,4 +1,3 @@
-oic==0.7.6
 pyjwkest==1.0.1
 cherrypy==3.2.4
 pyaml==15.03.1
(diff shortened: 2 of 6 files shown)


tpazderka:
Since the examples are SO broken, I am not against merging this with some stuff unfinished. Can you please update this to a current master as some stuff got already fixed?

claweyenuk:
I think this one is good (enough) to go now

claweyenuk:
The example README calls out #543 which I did not know how to work around

oidc_example/simple_rp/requirements.txt
```diff
@@ -1,4 +1,3 @@
-oic==0.7.6
 pyjwkest==1.0.1
```
tpazderka: This should be included in the installation and is probably extra.

oidc_example/simple_op/requirements.txt
```diff
@@ -1,6 +1,5 @@
 cherrypy==3.2.4
 pyaml==15.03.1
 Jinja2==2.7.3
-oic==0.7.6
 yubico-client==1.9.1
 pyjwkest==1.0.1
```
tpazderka: This should be included in the installation and is probably extra.

oidc_example/simple_rp/requirements.txt
claweyenuk: Should this be worded "should not be included"?

tpazderka: Yes, sorry. Applies to both comments.

tpazderka:
Merged manually.

PR #548, opened 2018-06-08
Deal with non-conformant JWKS.

rohe:
---

Recently encountered a very mal formed JWKS which made me realise that we didn't handle such
events very gracefully.


rohe:
The src/oic/oic/provider.py is just about wrapping lines at 80 chars. It just slipped in there was
not part of the changes I made intentional.
Specialised exception for malformed JWKS is a good idea.

PR #549, opened 2018-06-08
Bailed out early if 'iss' is missing from a signed JWT.

rohe:
---

A common error when people are constructing JWT's is to miss out on specifying 'iss'.
This will infallible lead to an exception saying no key matching the kid (if such a one was given)
could be found. When i reality the problem was that no 'iss' was given.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on the [KeepAChangeLog] project.
 ## 0.14.0 [2018-05-15]
 
 ### Fixed
+- [#549] Added a check for the presence of 'iss' when verifying the signature of a JWT
 - [#534] Fixed a bug in client_secret_basic authentication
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -45,6 +46,7 @@ The format is based on the [KeepAChangeLog] project.
 [#532]: https://github.com/OpenIDC/pyoidc/pull/532
 [#498]: https://github.com/OpenIDC/pyoidc/issues/498
 [#534]: https://github.com/OpenIDC/pyoidc/pull/534
+[#549]: https://github.com/OpenIDC/pyoidc/pull/XXX
```
schlenk: should have pull/549 here instead of pull/XXX

schlenk: This should also be moved to a different section, as 0.14.0 is already released.

rohe:
Covered by #577

PR #557, opened 2018-07-26
Make PKCE code generation RFC compliant

rectalogic:
https://tools.ietf.org/html/rfc7636

The code verifier is currently hexdigest before it is base64 encoded, this is incorrect.
pyoidc PKCE fails verification with all the providers I have tried it with (auth0, identityserver4).
Changing to not hexdigest (and also force ascii encoding per the spec, not whatever encoding the system is using) fixes PKCE.

- [X] The documentation has been updated, if necessary.
---


rectalogic:
Both client and provider changes are covered by existing tests in `tests/test_x_client.py` - test_pkce_verify_256 and test_pkce_verify_512

tpazderka:
The fail in verification is not tested and the basic implementation (not in extensions) is not covered as well.

tests/test_client.py
```diff
@@ -271,3 +286,42 @@ def test_valid_client_info(self):
...
+        authn_broker = AuthnBroker()
+        authn_broker.add("UNDEFINED", DummyAuthn(None, "username"))
+        _prov = Provider("as",
+                         session_db_factory('https://connect-op.heroku.com'), {},
+                         authn_broker, Implicit(), verify_client)
+
+        assert _prov.verify_code_challenge(cv, args['code_challenge'], 'S512') is True
+        resp = _prov.verify_code_challenge('XXX', args['code_challenge'])
+        assert isinstance(resp, Response)
+        assert resp.info()['status_code'] == 401
```
tpazderka: Missing newline, that is the reason for failing quality checks.

tpazderka:
Merged.

PR #563, opened 2018-08-17
Fixed uncaught exception from oic.provider.Provider.auth_init

### LINKED ISSUE #562: Uncaught exception from auth_init
`auth_init` method in Provider can cause an uncaught Exception.

This is sort of unexpected and the Exception should be caught and an error response should be returned instead.

tpazderka:
Close #562

---


tpazderka:
I have added a test to cover the change.

PR #567, opened 2018-09-11
Requests timeout

tpazderka:
---

### DIFF
--- a/src/oic/utils/keyio.py
+++ b/src/oic/utils/keyio.py
@@ -418,14 +418,18 @@ class KeyJar(object):
     """ A keyjar contains a number of KeyBundles """
 
     def __init__(self, verify_ssl=True, keybundle_cls=KeyBundle,
-                 remove_after=3600):
+                 remove_after=3600, timeout=5):
         """
         :param verify_ssl: Do SSL certificate verification
+        :param timeout: Timeout for requests library. Can be specified either as
+            a single integer or as a tuple of integers. For more details, refer to
+            ``requests`` documentation.
         :return:
         """
         self.spec2key = {}
         self.issuer_keys = {}
         self.verify_ssl = verify_ssl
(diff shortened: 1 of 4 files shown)


tpazderka:
The `copy`, `restore` and `import_jwks` doesn't seem to be operating with network connectivity.

I do agree that we should add the timeout to `load_keys`, I will add a commit for that.

tpazderka:
Ah, nevermind... I see your point.

src/oic/utils/keyio.py
```diff
@@ -782,12 +786,12 @@ def dump(self):
     def restore(self, info):
         for issuer, keys in info.items():
             self.issuer_keys[issuer] = [self.keybundle_cls(
-                keys, verify_ssl=self.verify_ssl)]
+                keys, verify_ssl=self.verify_ssl, timeout=self.timeout)]
 
     def copy(self):
         copy_keyjar = KeyJar(verify_ssl=self.verify_ssl)
```
schlenk: The new KeyJar should inherit the timeout too.

src/oic/utils/keyio.py
```diff
@@ -460,7 +464,7 @@ def add(self, issuer, url, **kwargs):
         if "/localhost:" in url or "/localhost/" in url:
             kc = self.keybundle_cls(source=url, verify_ssl=False, **kwargs)
```
schlenk: this should probably get the timeout too.

src/oic/utils/keyio.py
tpazderka: True.

src/oic/utils/keyio.py
tpazderka: I am not sure about this. Although looking at the condition this can reach the network if the url is poorly chosen, so it might be safer to add it.

PR #574, opened 2018-11-07
Fuzz content-type detection

bunjiboys:
---
Apply some fuzziness to the `Content-Type` check for JWKS responses, to allow for return values including encoding.

As an example, if the server returns `Content-Type: application/json;charset=utf-8` a warning is display about incorrect Content Type, despite this being a valid option.


tpazderka:
Just fix the quality build and this should be fine.

PR #577, opened 2018-11-16
Check that the issuer of a signed JWT exists in the Key Jar 

rohe:
Check that the issuer of a signed JWT exists in the Key Jar before any signature verification is attempted.

---

### DIFF
--- a/src/oic/oic/message.py
+++ b/src/oic/oic/message.py
@@ -10,6 +10,7 @@ import urllib
 
 import six
 from jwkest import jws
+from jwkest.jwt import JWT
 
 from oic.exception import InvalidRequest
 from oic.exception import IssuerMismatch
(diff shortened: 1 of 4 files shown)


tests/test_oic_message.py
```diff
@@ -529,17 +548,297 @@ def test_id_token():
...
+                 lifetime=3600)
+    _jws = packer.pack(**idt.to_dict())
+    msg = AuthorizationResponse(access_token=token2, id_token=_jws)
+    with pytest.raises(AtHashError):
+        verify_id_token(msg, check_hash=True, keyjar=kj,
+                        iss="https://sso.qa.7pass.ctf.prosiebensat1.com",
+                        client_id="554295ce3770612820620000")
+
+
+def test_verify_id_token_missing_at_hash():
```
tpazderka: There should be a test for missing `c_hash` as well.

src/oic/oic/message.py
```diff
@@ -258,6 +259,50 @@ def check_char_set(string, allowed):
...
+
+    _jws = str(instance["id_token"])
+    _packer = JWT()
+    _body = _packer.unpack(_jws).payload()
+    if 'keyjar' in kwargs:
+        try:
+            if _body['iss'] not in kwargs['keyjar']:
+                raise ValueError('Unknown issuer')
+        except KeyError:
+            pass
```
tpazderka: What is the purpose of this pass? What are we supposed to do if the `iss` is missing in the message?

This piece of code is also untested.

tests/test_oic_message.py
rohe: Added one

src/oic/oic/message.py
rohe: Dealt with this.

tests/test_oic_message.py
```diff
@@ -529,17 +548,371 @@ def test_id_token():
...
+                     'dYMmrcQksKaPkhdgRNYk3zzh5l7ewdDJ', ['sig'])
+    packer = JWT(kj, sign_alg='HS256', lifetime=3600,
+                 iss='https://example.com/op')
+    _jws = packer.pack(**idt.to_dict())
+    msg = AuthorizationResponse(id_token=_jws)
+    with pytest.raises(ValueError):
+        verify_id_token(msg, check_hash=True, keyjar=kj,
+                        iss="https://sso.qa.7pass.ctf.prosiebensat1.com",
+                        client_id="554295ce3770612820620000")
+
```
tpazderka: An extra blank line.

PR #580, opened 2018-11-28
Fix UnicodeDecodeError in verify_redirect_uris

### LINKED ISSUE #579: UnicodeDecodeError in verify_redirect_uri
The code in `oic.provider.Provider.verify_redirect_uris` fails with `UnicodeDecodeError` if the `redirect_uri` in request contains unicode characters.

This is due to a bug in `future.backports.urllib.parse.urlparse`, the best solution is to use `urlparse` from `future.moves`.

tpazderka:
Close #579

---


tests/test_oic_provider.py
```diff
@@ -1051,6 +1051,14 @@ def test_verify_redirect_uris_with_non_https_redirect_uri_implicit_flow(
 
         assert str(exc_info.value) == "None https redirect_uri not allowed"
 
+    def test_verify_redirect_uris_unicode(self):
+        params = {"application_type": "web",
+                  "redirect_uris": ["http://example.com/aůťhž"],
```
schlenk: At least for python 2 i would prefer the \u or even UTF-8 bytes version of the string in the test.
Otherwise this is a bit random under Python 2, depending on system encoding.

tpazderka: True, changed.

PR #590, opened 2019-01-15
Client id in cookie

tpazderka:
---


tpazderka:
The only changes apart from your PR @rohe is the separation of getting the data from cookie and an extra test with a wrong cookie to get full coverage on testing.

Otherwise it should be the same (small quality fixes).

Issue #341, opened 2017-05-04
oic.utils.authn.user.BasicAuthn.verify_password should use constant time check


schlenk:
The verify password method uses a naive == comparision for password strings, which allows timing attacks.

It should use a constant time comparision.

As it is a dependency anyway, maybe use:
cryptography.hazmat.primitives.constant_time.bytes_eq

https://cryptography.io/en/latest/hazmat/primitives/constant-time/

lwm:
Fair point. Would appreciate a PR @schlenk! You seem to know what you're talking about :smile:

schlenk:
Thinking about it. It should probably not do the password handling itself at all but use a decent library like [passlib](https://pypi.python.org/pypi/passlib) to handle encrypted password storage and comparision. 

At least if the code is more than a glorified example.

lwm:
I'd be OK with adding this `passlib` dependency! Go for it :+1:

PR #595, opened 2019-01-24
Drop py2

### LINKED ISSUE #578: Drop support for python 2
I think that it is about time we drop support for python 2. The remaining lifetime is only one year and dropping this support would allow some much needed code cleanup.

We should do a last release which would support both versions of python towards the end of the year and then drop python 2 entirely in 2019.

If we agree on this, I can focus on that.

tpazderka:
---

Close #578

### DIFF
--- a/src/oic/utils/authn/user.py
+++ b/src/oic/utils/authn/user.py
@@ -1,4 +1,5 @@
 import base64
+import hmac
 import logging
 import time
 from urllib.parse import parse_qs
@@ -356,7 +357,7 @@ class BasicAuthn(UserAuthnMethod):
     def verify_password(self, user, password):
         if user in self.passwd:
             _pwd = self.passwd[user]
-            if _pwd != password:
+            if not hmac.compare_digest(_pwd.encode(), password.encode()):
                 raise FailedAuthentication('Wrong user/password combination')
         else:
             raise FailedAuthentication('Wrong user/password combination')
--- a/src/oic/utils/authn/user.py
+++ b/src/oic/utils/authn/user.py
@@ -1,15 +1,12 @@
-# coding=utf-8
-from future.backports.urllib.parse import unquote_plus
-from future.backports.urllib.parse import urlencode
-from future.backports.urllib.parse import urlsplit
-from future.backports.urllib.parse import urlunsplit
-from future.moves.urllib.parse import parse_qs
-
 import base64
 import logging
 import time
+from urllib.parse import parse_qs
+from urllib.parse import unquote_plus
+from urllib.parse import urlencode
+from urllib.parse import urlsplit
+from urllib.parse import urlunsplit
 
-import six
 from jwkest import as_unicode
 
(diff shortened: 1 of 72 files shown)


tpazderka:
> @tpazderka Is there any kind of warning (or exception) being raised for cases where a user attempts to use this version with Python2?

If you try to run the code from source files, they would raise an ImportError due to moved libraries.

The support for Python2 is not marked in setup.py, but I guess we can add a check in setup for python version and produce a readable warning.

Kilo59:
> The support for Python2 is not marked in setup.py, but I guess we can add a check in setup for python version and produce a readable warning.

✔️ 

I do think this is a bit of an edge case.
If someone is installing directly from source they probably know enough to figure out what's going on.
But occasionally I will make a mistake and use the wrong version of python and it's not always immediatley apparent why I'm getting an import or syntax error for example.

src/oic/utils/http_util.py
```diff
@@ -284,29 +275,29 @@ def cookie_signature(key, *parts):
...
+       :type sig: str
        :param key: The HMAC key to use.
        :type key: bytes
        :param parts: List of parts to include in the MAC
        :type parts: list of bytes or strings
        :raises: `InvalidCookieSign` when the signature is wrong
     """
-    assert isinstance(sig, text_type)
+    assert isinstance(sig, str)
     return safe_str_cmp(sig, cookie_signature(key, *parts))
```
schlenk: safe_str_cmp was only needed because ancient Py 2.7 versions have no hmac.compare_digest() for constant time comparision.

So this can be replaced by hmac.compare_digest().

tpazderka: OK, I will probably change that as well.

I have later noticed some other checks for python version, so I give it another look.

We are not rushing to merge this, I still need to make the bugfix released before I am going to merge this anyway :)

schlenk: ok. Had the impression you were rather eager to merge it as soon as possible ;-).

tpazderka: I want to merge before doing more changes to the codebase as it should make the work easier :)

src/oic/utils/authn/user.py
```diff
@@ -359,9 +356,6 @@ def __init__(self, srv, pwd, ttl=5):
     def verify_password(self, user, password):
         if user in self.passwd:
             _pwd = self.passwd[user]
-            if six.PY2:
-                if isinstance(_pwd, str):
-                    _pwd = _pwd.decode('utf-8')
             if _pwd != password:
```
schlenk: As this is Py3 now, we have a constant time compare (hmac.compare_digest()).

So maybe kill https://github.com/OpenIDC/pyoidc/issues/341 too by using it to compare the passwords.

tpazderka: Good catch :) Will do that.

tpazderka:
Just a note to myself... This breaks the certification tests... Will need to investigate before we can merge this.

PR #599, opened 2019-01-29
move alabaster to docs extras

### LINKED ISSUE #598: alabaster should not be in runtime deps
Hi, ✋ 

As I can see the `alabaster` dependency used only for documentation in https://github.com/OpenIDC/pyoidc/blob/master/doc/conf.py#L1.
I think this dependency needed only for developers, not in install deps https://github.com/OpenIDC/pyoidc/blob/master/setup.py#L88

artofhuman:
---

Closes #598


tpazderka:
Merged manually.

PR #604, opened 2019-02-08
Cleanup

tpazderka:
---
Some code cleanup of unused variables.

Should be safe :)


tpazderka:
> I'm not always a fan of those _ replacements, as they can make reading the code harder when the result types are not fully memorized. But well, dead code tends to be worse.
> 
> Some of the uses where lots of _ are used in splitting a type into parts could probably benefit from named tuples or similar self documenting types instead of plain tuples. e.g. the _spec.items() values in message.py.

Yes, I agree and I am planning to use named tuples for _spec definitions. I arrived to the same conclusion as you :) Will focus on this conversion next.

PR #605, opened 2019-02-09
Use namedtuple for c_param definitions

tpazderka:
---

### DIFF
--- a/src/oic/oauth2/message.py
+++ b/src/oic/oauth2/message.py
@@ -152,6 +152,18 @@ class Message(MutableMapping):
         for key, val in self.c_default.items():
             self._dict[key] = val
 
+    @staticmethod
+    def _extract_cparam(key, _spec):
+        """Extract ParamDefinition for a given key.
+
+        The key can be direct attribute or lang typed attribute.
+        If ParamDefinition is not found, tries to return "*" attribute, if it exists, otherwise returns None.
+        """
+        for _key in (key, key.split('#')[0], '*'):
+            if _key in _spec:
+                return _spec[_key]
+        return None
+
     def to_urlencoded(self, lev=0):
         """
         Creates a string using the application/x-www-form-urlencoded format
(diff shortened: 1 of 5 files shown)


src/oic/oauth2/message.py
```diff
@@ -160,26 +161,32 @@ def to_urlencoded(self, lev=0):
...
-                if req and attribute not in self._dict:
+            for attribute, cparam in _spec.items():
+                if cparam.required and attribute not in self._dict:
                     raise MissingRequiredAttribute("%s" % attribute,
                                                    "%s" % self)
 
         params = []
 
         for key, val in self._dict.items():
             try:
```
schlenk: The whole nested try/except could be done like this:

    cparam = _spec.get(key, _spec.get(key.split('#')[0], _spec.get('*'))

Not sure if thats any better.

Nearly the same pattern (just differs in the finally part) happens in line 261 and 310, and very similar in line 355, just written differently  so maybe a small helper function would be good?
e.g.

    def extract_cparam(_spec, key):
        for _key in (key, key.split('#')[0], '*'):
            if _key in _spec:
                return _spec[key]

src/oic/oauth2/message.py
```diff
@@ -711,30 +722,30 @@ def verify(self, **kwargs):
         except KeyError:
             _allowed = {}
 
-        for (attribute, (typ, required, _, _, na)) in _spec.items():
+        for (attribute, cparam) in _spec.items():
```
schlenk: the ( ) around the tuple are not necessary

src/oic/oauth2/message.py
```diff
@@ -711,30 +722,30 @@ def verify(self, **kwargs):
...
                 continue
 
             try:
                 val = self._dict[attribute]
             except KeyError:
-                if required:
+                if cparam.required:
                     raise MissingRequiredAttribute("%s" % attribute)
                 continue
             else:
```
schlenk: The else is not needed, all pathes in the except clause either raise or continue.

src/oic/oauth2/message.py
```diff
@@ -711,30 +722,30 @@ def verify(self, **kwargs):
...
-                if typ == bool:
+                if cparam.type == bool:
```
schlenk: can be simplified to:

    if not cparam.type == bool and not val:
        if cparam.required:
            raise...
        continue

src/oic/oauth2/message.py
```diff
@@ -711,30 +722,30 @@ def verify(self, **kwargs):
...
 
             if attribute not in _allowed:
                 continue
 
-            if isinstance(typ, tuple):
+            if isinstance(cparam.type, tuple):
                 _ityp = None
-                for _typ in typ:
+                for _typ in cparam.type:
                     try:
```
schlenk: I doubt this can raise a ValueError. Not sure though. The entire check is a little weird looking.
Why is this not part of type_check() itself?

If the cparam.type cannot contain a None value, this can be simplified as:

    for _typ in cparam.type:
         try:
             self._type_check(...)
             break
         except ValueError:
             pass
    else:
          raise NotAllowedValue(val)

tpazderka:
> The change looks fine, just spotted some weirdness around the changes.

Yeah, I have noticed some of the things you are mentioning. I wasn't sure if this is a good opportunity to do a refactor, but maybe it is. I will give it a go and change some things...

schlenk:
> 
> 
> > The change looks fine, just spotted some weirdness around the changes.
> 
> Yeah, I have noticed some of the things you are mentioning. I wasn't sure if this is a good opportunity to do a refactor, but maybe it is. I will give it a go and change some things...

Whatever feels right. The change with the parameter type just made the duplication more visible.

PR #606, opened 2019-02-10
Integration tests

tpazderka:
---


tpazderka:
i am going to ignore the rest of the Codacy issues as they do not make much sense. We are not worried about pinning the dependencies and we have to do two runs of apt-get install so there is no sense in clearing the cache in-between the runs.

docker/op/cert.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN CERTIFICATE-----
```
schlenk: I usually prefer to just create a self signed cert on the fly. This one is only valid till Oct 2019, so would this work after that date too?

tpazderka: OK, I will make a separate issue for that.

schlenk: yes, sounds fine. For testing its not really an issue.

PR #612, opened 2019-02-17
Mypy

tpazderka:
---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -9,6 +9,7 @@ import socket
 import sys
 import time
 import traceback
+import warnings
 from functools import cmp_to_key
 from urllib.parse import parse_qs
 from urllib.parse import splitquery  # type: ignore
@@ -251,8 +252,8 @@ class Provider(AProvider):
         for endp in self.endp:
             if endp.etype == 'registration':
                 endpoint = urljoin(self.baseurl, endp.url)
-                DeprecationWarning("Using `register_endpoint` is deprecated, please use "
-                                   "`registration_endpoint` instead.")
+                warnings.warn("Using `register_endpoint` is deprecated, please use `registration_endpoint` instead.",
+                              DeprecationWarning)
                 self.register_endpoint = endpoint
                 break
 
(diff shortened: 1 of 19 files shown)


tpazderka:
Hm, the Py34 on Windows is failing for some reason. I am considering dropping the support for Py34 as its EOL is in about a month...

@schlenk Any objections?

schlenk:
@tpazderka I don't have objections about dropping Python 3.4 alongside the Python 2.7 drop, its basically the same reasoning.

tpazderka:
OK, 3.4 removed as well.

src/oic/oic/provider.py
```diff
@@ -247,9 +247,12 @@ def __init__(self, name, sdb, cdb, authn_broker, userinfo, authz,
         self.extra_claims = extra_claims
         self.extra_scope_dict = extra_scope_dict
 
+        self.register_endpoint = None
         for endp in self.endp:
             if endp.etype == 'registration':
                 endpoint = urljoin(self.baseurl, endp.url)
+                DeprecationWarning("Using `register_endpoint` is deprecated, please use "
```
schlenk: Why only create a `DeprecationWarning` instance? Shouldn't it be used with warning.warn(...) ?

src/oic/utils/http_util.py
```diff
@@ -546,14 +546,15 @@ def wsgi_wrapper(environ, start_response, func, **kwargs):
...
 
-    def setServer(self, server):
+    @srv.setter
+    def srv(self, server):
         self._srv = server
 
-    srv = property(getServer, setServer)
-
     def __init__(self, srv, ttl=5, secure=True, httponly=True):
         self.srv = None
```
schlenk: Why keep this when having the property above?

src/oic/oic/provider.py
tpazderka: Yeah, dumb mistake :)

src/oic/utils/http_util.py
tpazderka: True

Issue #608, opened 2019-02-14
Get the ID token in JWT format


layoaster:
The response I got from the `do_access_token_request` contains all tokens but the ID token gets decoded. This is really useful but is there a way to get the original JWT (ID Token) sent from the OIDC server?

Regards,

schlenk:
You can wrap the response_cls to get it:

```python
     from oic.oauth2.message import AccessTokenResponse
     class ResponseWrapper(AccessTokenResponse):
            """Wrap Response to capture encoded id_token"""
            def __init__(self, *args, **kwargs):
                super(ResponseWrapper, self).__init__(*args, **kwargs)
                self.raw_id_token = None

            def verify(self, **kwargs):
                if "id_token" in self:
                    self.raw_id_token = self['id_token']
                return super(ResponseWrapper, self).verify(**kwargs)
        response_cls = ResponseWrapper

        resp = client.do_access_token_request(
            state=aresp["state"], request_args=args,
            authn_method=authn_method,
            keyjar=client.keyjar,
            response_cls=response_cls)
```

layoaster:
Thanks, that solved the question but afterwards, making requests to the userinfo endpoint fails:
```
 user_info = oidc_client.do_user_info_request(state=auth_resp["state"])
```

Error traceback:
```
web-client_1  |   File "/code/app.py", line 124, in callback
web-client_1  |     user_info = oidc_client.do_user_info_request(state=auth_resp["state"])
web-client_1  |   File "/usr/local/lib/python3.7/site-packages/oic/oic/__init__.py", line 871, in do_user_info_request
web-client_1  |     scope, **kwargs)
web-client_1  |   File "/usr/local/lib/python3.7/site-packages/oic/oic/__init__.py", line 805, in user_info_request
web-client_1  |     if token.is_valid():
web-client_1  | AttributeError: 'NoneType' object has no attribute 'is_valid'
```

Debug logs from library:
```
[...]

PR #613, opened 2019-02-18
Fix: Grant storage for access tokens response subclasses

layoaster:
---
It fixes the issue we discovered on #608

### DIFF
--- a/tests/test_oic.py
+++ b/tests/test_oic.py
@@ -34,6 +34,7 @@ from oic.oic.message import OpenIDSchema
 from oic.oic.message import RefreshAccessTokenRequest
 from oic.oic.message import RefreshSessionRequest
 from oic.oic.message import RegistrationRequest
+from oic.oic.message import SINGLE_OPTIONAL_STRING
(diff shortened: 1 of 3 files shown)


tests/test_oic.py
```diff
@@ -143,6 +144,41 @@ def test_access_token_request(self):
         assert _eq(resp.keys(),
                    ['token_type', 'state', 'access_token', 'scope'])
 
+    def test_access_token_request_with_custom_response_class(self):
```
tpazderka: This test should probably go to `test_oauth2_consumer` since the fix is there.

layoaster: Ok, I placed it at the `oic` level since `oauth2` is the underliying infrastructure so everything get tested altogether.  If you don't mind, I'm going to leave the test in `oic` and make another one in `test_oauth2_consumer`.

tpazderka: The reasoning behind my comment was that the test should be where the change was.

But it sort of makes sense to test both, since we want to make sure that the oic doesn't overload completely.

PR #621, opened 2019-02-27
Make OP2 Great Again!

mblomdahl:
---

This PR marks an initial step towards making the OIDC OP2 example self-contained and usable by a noob like me (and friends). Following the instructions in the new README should be enough to get something up and running, which was pretty-damn-hard for someone who doesn't know anything about OICD initially. This change is mainly intended to get new (future) contributers onboard and spur interest. Probably not so interesting for all of you OIDC pros out there... :)


oidc_example/op2/requirements.txt
```diff
@@ -0,0 +1,19 @@
+alabaster==0.7.12
```
tpazderka: I am not a fan of having the dependencies pinned in requirements file. These should only list first level dependencies without specific versions.

The specific versions should rather be in constraints file.

oidc_example/op2/server.py
```diff
@@ -373,25 +366,33 @@ def application(self, environ, start_response):
...
-    parser.add_argument('-d', dest='debug', action='store_true')
-    parser.add_argument('-p', dest='port', default=80, type=int)
-    parser.add_argument('-t', dest='tls', action='store_true')
-    parser.add_argument('-k', dest='insecure', action='store_true')
-    parser.add_argument(
-        '-c', dest='capabilities',
-        help="A file containing a JSON representation of the capabilities")
-    parser.add_argument('-i', dest='issuer', help="issuer id of the OP",
+    parser.add_argument('-v', '--verbose', dest='verbose', action='store_true',
+                        help='Moar verbose output')
```
tpazderka: Typo, `More`

oidc_example/op2/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+    git clone https://github.com/OpenIDC/pyoidc.git
+    cd pyoidc/oidc_example/op2/
+    python3 -m venv venv && . venv/bin/activate
+    pip install -r requirements.txt
+
+
+### Client Management
+
+To be able to start up the project you have to create a client in
```
tpazderka: Can you please reformat to one sentence per line? I think that keeps the file more readable.

mblomdahl: Fixed in 81d56dc!

oidc_example/op2/server.py
mblomdahl: Corrected!

oidc_example/op2/requirements.txt
mblomdahl: Interesting concept, have not tired the constraints file before. :) I gave it a stab in my fix-remarks commit here: 81d56dc

Let me know what you think!

mblomdahl:
Sqashed all the scrappy back-and-forth fiddling commits and force-pushed. Content wise, everything before commit 81d56dc is identical to the initial PR scope.

oidc_example/op2/requirements.txt
tpazderka: Yes, this is exactly what I want to see. The difference here is that the constraints are only applied when you want and also they do not force the dependencies to be installed, so if one of the dependencies drops something, we no longer force the installation of it.

oidc_example/op2/README.md
tpazderka: I see what you did there ... :)

oidc_example/op2/server.py
tpazderka: I see what you did there :)

tpazderka:
Thanks for the update. I will play with the OP2 a little bit before I merge this. I will let you know if I have more comments after that.

PR #636, opened 2019-03-31
Fix ROPC in extensions

### LINKED ISSUE #615: Usage Question: ROPCAccessToken usage causes assert req["grant_type"] == "authorization_code" assert error
Hi, I'm trying to get the Resource Ownger Password Credentials Grant working and i'm getting an error. When i'm trying to run the following code 
 
        keys = [{ 
                'type': 'RSA',
                'key':'../OP/keys/key.pem',
                'use': ['enc', 'sig']
        }]

        _, keyjar, _ = build_keyjar(keys)
        global client
        client = Client(verify_ssl=False, keyjar=keyjar, client_authn_method=CLIENT_AUTHN_METHOD)
        issuer_url = client.wf.discovery_query('localhost')
        provider_info = client.provider_config(issuer_url)

        reg_info = {'application_type': 'native',
                    'application_name': 'Example RP',
                    'redirect_uris': ['http://127.0.0.1:5001/HybFlow'],
                    }

        client_reg = client.register(provider_info["registration_endpoint"], **reg_info)
        client.store_registration_info(client_reg)

        request_args = {
            'grant_type': 'password',
            'client_id': client.client_id,
            'username': 'diana',
            'password': 'krall'
        }

       
        client.do_access_token_request(request=ROPCAccessTokenRequest,request_args=request_args)                    
        return

I get the following assert error. Can anyone please guide me on what i am doing wrong in trying to get a token using the Password Grant Flow?

`
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask_restful\__init__.py", line 269, in error_router
    return original_handler(e)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\_compat.py", line 34, in reraise
    raise value.with_traceback(tb)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask_restful\__init__.py", line 269, in error_router
    return original_handler(e)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\_compat.py", line 34, in reraise
    raise value.with_traceback(tb)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask_restful\__init__.py", line 458, in wrapper
    resp = resource(*args, **kwargs)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask\views.py", line 88, in view
    return self.dispatch_request(*args, **kwargs)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\flask_restful\__init__.py", line 573, in dispatch_request
    resp = meth(*args, **kwargs)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\src\OP.py", line 102, in post
    authn=request.headers.environ['HTTP_AUTHORIZATION']
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\oic\oic\provider.py", line 1115, in token_endpoint
    return self._access_token_endpoint(req, **kwargs)
  File "C:\Users\bogdan\Projects\Learning\Python\oidc\OP\venv\lib\site-packages\oic\oic\provider.py", line 964, in _access_token_endpoint
    assert req["grant_type"] == "authorization_code"
AssertionError
`

tpazderka:
Close #615

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#607] Fixed key recovery on encryption of payload
 - [#618] Prettified `client_management.py` CLI and wrapped it as
          a setup.py console script `oic-client-management`
+- [#615] Fix ROPC grant in the extensions provider
 
 ### Changed
 - [#578] Dropped python 2.7 support
@@ -40,6 +41,7 @@ The format is based on the [KeepAChangeLog] project.
 [#56]: https://github.com/OpenIDC/pyoidc/issues/56
 [#624]: https://github.com/OpenIDC/pyoidc/pull/624
 [#629]: https://github.com/OpenIDC/pyoidc/issues/629
+[#615]: https://github.com/OpenIDC/pyoidc/issues/615
 
 ## 0.15.1 [2019-01-31]
 
--- a/src/oic/extension/provider.py
+++ b/src/oic/extension/provider.py
@@ -54,6 +54,7 @@ from oic.utils.keyio import KeyBundle
 from oic.utils.keyio import KeyJar
 from oic.utils.sanitize import sanitize
 from oic.utils.sdb import AccessCodeUsed
+from oic.utils.sdb import AuthnEvent
 from oic.utils.time_util import utc_time_sans_frac
 from oic.utils.token_handler import NotAllowed
 from oic.utils.token_handler import TokenHandler
@@ -640,19 +641,30 @@ class Provider(provider.Provider):
         return Response(atr.to_json(), content="application/json")
 
     def password_grant_type(self, areq):
-        _at = self.token_handler.get_access_token(areq['client_id'],
-                                                  scope=areq['scope'],
-                                                  grant_type='password')
-        _info = self.token_handler.token_factory.get_info(_at)
-        try:
-            _rt = self.token_handler.get_refresh_token(
-                self.baseurl, _info['access_token'], 'password')
-        except NotAllowed:
-            atr = self.do_access_token_response(_at, _info, areq['state'])
-        else:
-            atr = self.do_access_token_response(_at, _info, areq['state'], _rt)
+        """
+        Token authorization using Resource owner password credentials.
 
-        return Response(atr.to_json(), content="application/json")
+        RFC6749 section 4.3
+        """
+        # `Any` comparison tries a first broker, so we either hit an IndexError or get a method
+        try:
+            authn, authn_class_ref = self.pick_auth(areq, "any")
+        except IndexError:
+            err = TokenErrorResponse(error='invalid_grant')
+            return Unauthorized(err.to_json(), content='application/json')
+        identity, _ts = authn.authenticated_as(username=areq['username'], password=areq['password'])
+        if identity is None:
+            err = TokenErrorResponse(error='invalid_grant')
+            return Unauthorized(err.to_json(), content='application/json')
+        # We are returning a token
+        areq['response_type'] = ['token']
+        authn_event = AuthnEvent(identity["uid"], identity.get('salt', ''),
+                                 authn_info=authn_class_ref,
+                                 time_stamp=_ts)
+        sid = self.setup_session(areq, authn_event, self.cdb[areq['client_id']])
+        _at = self.sdb.upgrade_to_token(self.sdb[sid]['code'], issue_refresh=True)
+        atr = AccessTokenResponse(**by_schema(AccessTokenResponse, **_at))
+        return Response(atr.to_json(), content="application/json", headers=OAUTH2_NOCACHE_HEADERS)
 
     def refresh_token_grant_type(self, areq):
         at = self.token_handler.refresh_access_token(
--- a/tests/test_x_provider.py
+++ b/tests/test_x_provider.py
@@ -16,6 +16,7 @@ from oic.oauth2.message import AccessTokenRequest
 from oic.oauth2.message import AccessTokenResponse
 from oic.oauth2.message import AuthorizationRequest
 from oic.oauth2.message import AuthorizationResponse
+from oic.oauth2.message import ROPCAccessTokenRequest
 from oic.oauth2.message import TokenErrorResponse
 from oic.utils.authn.authn_context import AuthnBroker
 from oic.utils.authn.client import verify_client
@@ -118,16 +119,21 @@ def test_lv_pack_unpack():
 
 
 class DummyAuthn(UserAuthnMethod):
-    def __init__(self, srv, user):
+    def __init__(self, srv, user, password=None):
         UserAuthnMethod.__init__(self, srv)
         self.user = user
+        self.password = password
 
     def authenticated_as(self, cookie=None, **kwargs):
+        if self.password is not None and kwargs.get('password') != self.password:
+            return None, 0
         return {"uid": self.user}, time.time()
 
 
 AUTHN_BROKER = AuthnBroker()
 AUTHN_BROKER.add("UNDEFINED", DummyAuthn(None, "username"))
+AUTHN_BROKER2 = AuthnBroker()
+AUTHN_BROKER2.add("UNDEFINED", DummyAuthn(None, "username", "password"))
 # dealing with authorization
 AUTHZ = Implicit()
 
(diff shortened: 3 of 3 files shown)


src/oic/extension/provider.py
```diff
@@ -640,19 +640,9 @@ def client_credentials_grant_type(self, areq):
...
-        try:
-            _rt = self.token_handler.get_refresh_token(
-                self.baseurl, _info['access_token'], 'password')
-        except NotAllowed:
-            atr = self.do_access_token_response(_at, _info, areq['state'])
-        else:
-            atr = self.do_access_token_response(_at, _info, areq['state'], _rt)
-
-        return Response(atr.to_json(), content="application/json")
+        _at = self.sdb.upgrade_to_token(key=areq['password'], issue_refresh=True)
```
schlenk: This looks like it would issue a refresh token to a public client that may not have 'offline_access' scope.

Probably needs some scope validation too (like it is done in oic.provider for the code_grant_type to determine if a refresh token should be issued.)

tpazderka: I am not 100% sure on this. The specs are a little bit vague in this respect and there is no mention of the scoping https://tools.ietf.org/html/rfc6749#section-4.3.3.

The issuing of the access credentials is entirely out of scope of the specs, so I am not sure where the `offline_access` scope would come in. We can add a flag to the ClientDB just for this purpose.

src/oic/extension/provider.py
```diff
@@ -640,19 +640,9 @@ def client_credentials_grant_type(self, areq):
         return Response(atr.to_json(), content="application/json")
 
     def password_grant_type(self, areq):
-        _at = self.token_handler.get_access_token(areq['client_id'],
```
schlenk: I don't see any kind of password / username check in here.

As we don't have a real user database to ask, there should probably be a call to some authn code somewhere in here, but i don't know how it would look. Probably just some callback to override by users of the class when they attach some form of authentication. 

The existing authn code isn't entirely useful here, as we do not know which authn method to pick (e.g. it might be a local password database or an LDAP based one, or both in parallel under different ACRs.), and we do not know how to pass username, password as the `authenticated_as()` methods do not have a standard for that.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#607] Fixed key recovery on encryption of payload
 - [#618] Prettified `client_management.py` CLI and wrapped it as
          a setup.py console script `oic-client-management`
+- [#615] Fir ROPC grant in the extensions client
```
schlenk: Fix ;-)

And isn't it the extensions provider?

src/oic/extension/provider.py
tpazderka: The credentials check are done in the `token_endpoint` method. This just handles the specifics. So we already have a verified client and we are just creating an access token based on that.

The user authentication is separate and was done in previous step.

src/oic/extension/provider.py
schlenk: See https://tools.ietf.org/html/rfc6749#section-4.3.2 where the request is described. The rules for offline_access scope would come from https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess.

But you are right, the spec is vague and the OpenID connect thing mentions a MAY for issuing the refresh token anyway.

I would pick a default policy of deny the refresh token and maybe add a configurable option to grant it anyway, if configured that way.

tpazderka: Yes, the extension is OAuth2 only. My understanding of the specs is that we should either issue or not issue the refresh but not distinguish between clients.

CHANGELOG.md
tpazderka: Fixed

src/oic/extension/provider.py
tpazderka: And as for the password in the request, I am using a "trick" with session identifier. The establishing of username/password is out of scope of the specs, and it would need to be handled manually anyway, so I am using the existing sessionDB there. Not sure if any way for easy customization is really necessary.

src/oic/extension/provider.py
schlenk: Ok, that invalidates my comment about offline_access scope.

The spec is vague, but hints at some distinction between clients in section 2.1.

But ok, issuing a refresh_token seems fine in this case.

src/oic/extension/provider.py
schlenk: I don't really see where the user authn happens in this case. Client authn happens in the token_endpoint, yes. 

But for the token_endpoint() no authn based on the user happens anywhere, at least not in the example code that uses it (https://github.com/OpenIDC/pyoidc/blob/master/oauth_example/as/as.py) that goes straight from request to the provider.token_endpoint().

Your trick obviously works for a trivial demo. But how would you add ROCP to the example above, e.g. to log in with 'diana'?

src/oic/extension/provider.py
tpazderka: OK, we will keep it this way then.

src/oic/extension/provider.py
tpazderka: As is said in 4.3.1, the method of obtaining the username/password is out of scope. The user authentication would be done there and the client credentials issued there as well.
This would require the user to rewrite/augment AuthorizationEndpoint to fit this job.

I guess that this brings an interesting question - what should we do with the extension module? We can probably either drop it completely (we don't need the code for OIDC) or integrate it into Oauth module (we will try to be a complete OAuth 2.0 implementation).

schlenk: 4.3.1 talks about the client. The flow in 4.3 (C) talks about the server, which is the provider, and 4.3.2 explicitly says "The authorization server MUST:  
- require client authentication for confidential clients or for any client that was issued client credentials (or with other authentication requirements),
- authenticate the client if client authentication is included, and
- validate the resource owner password credentials using its existing password validation algorithm.
"
We do the first two here, but skip the third. I would be ok if the code called a method like 'verify_user_and_password' that simply raises NotImplementedError() in the default. It should return the uid of the authenticated user (a bit like the authn plugins). A user would need to add that part.

There are some interesting things in the extension module, that would be nice to have (more features for client registration for example, or the software statement/code stuff) so integrating it would probably be the best longer term thing to do.

tpazderka: Ah, I see. I think I misunderstood the specs. Will update accordingly.

tpazderka:
OK, so I re-read the specs and now I see your point :) Sorry that it took me so long, I misunderstood the Authorization endpoint comment in the specs and it never occurred to me that the username/password are of the user...

PR #643, opened 2019-04-04
Re-add symmetric keys if issuer keys are missing

### LINKED ISSUE #639: Provider looses symmetric key of dynamic clients on restart
When a client registers dynamically and requests encryption, this adds a key dervived from the client secret to the keyjar. This key will be used to encrypt userinfo and id_token responses, if the client asked for symmetric encryption. For RSA the jwks_uri provides the keys, so that works.

This is defined in:
https://openid.net/specs/openid-connect-core-1_0.html#Encryption (Symmetric Encryption)

But as the keyjar has no persistence option, that symmetric key is lost during a restart, so a client trying to authenticate after an OP restarts, causes an failure to encrypt.

The key should be re-added to the keyjar as needed, as it is done during client registration:
https://github.com/OpenIDC/pyoidc/blob/f4781ba38fad0f299ca7d17fcff42a5ca3fdd00d/src/oic/oic/provider.py#L1555

Or the keyjar needs a persistence option...

tpazderka:
Close #639

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on the [KeepAChangeLog] project.
          a setup.py console script `oic-client-management`
 - [#615] Fix ROPC grant in the extensions provider
 - [#640] Use more secure random generator for client_secret
+- [#639] Make sure symmetric keys are available after server restart
 
 ### Changed
 - [#578] Dropped python 2.7 support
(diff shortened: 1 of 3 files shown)


tpazderka:
Added tests with `userinfo`, I will wait for the CI before merging.

PR #647, opened 2019-04-12
Use black for code formatting

tpazderka:
---


tpazderka:
Codacy is just looking at the changes and getting confused.

The failing integration is because of expired key in google repo, hopefully that will be sorted shortly.

tpazderka:
I was thinking the same about the function annotations. The important thing is that the code should now be consistent which I think is a plus :)

Issue #638, opened 2019-04-02
Move methods from extensions


tpazderka:
Keeping the base classes and the extension classes in sync is proving to be difficult, it might be a good idea to pull the extension code to the main oauth module.

The most interesting code is:
* Client
 - [ ] do_client_read_request
 - [ ] do_client_registration
 - [ ] do_client_update_request
 - [ ] do_client_delete_request
 - [ ] authorization with code challenge
 - [ ] do_token_revocation
 - [ ] do_token_introspection

* Server
 - [ ] client_info
 - [ ] do_client_registration
 - [ ] client_info_update
 - [ ] verify_code_challenge
 - [ ] client_credentials_code_grant
 - [ ] password_grant_type
 - [ ] refresh_token_grant_type
 - [ ] revocation_endpoint
 - [ ] introspection_endpoint

Things to note:
* Some of these methods are already implemented in `oidc` module and it might be possible to merge these. 
* Some of these methods are broken and will need to be rewritten.
* Some of these methods are OAuth2.0 specific and will need to be disabled in `oic` module.

PR #648, opened 2019-04-12
Moved provider_info_endpoint from extensions to oauth

tpazderka:
Part of #638

---

### DIFF
--- a/src/oic/oauth2/provider.py
+++ b/src/oic/oauth2/provider.py
@@ -1,15 +1,22 @@
 #!/usr/bin/env python
+import copy
 import hashlib
 import logging
 import os
 import sys
 import traceback
 import warnings
+from functools import cmp_to_key
+from typing import Dict  # noqa - This is used for MyPy
+from typing import List  # noqa - This is used for MyPy
+from typing import Union  # noqa - This is used for MyPy
 from urllib.parse import parse_qs
 from urllib.parse import splitquery  # type: ignore
 from urllib.parse import unquote
 from urllib.parse import urljoin
 from urllib.parse import urlparse
+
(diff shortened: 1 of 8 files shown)


src/oic/oic/provider.py
```diff
@@ -1786,11 +1777,9 @@ def delete_registration(self, authn, request, **kwargs):
 
     def create_providerinfo(self, pcr_class=None, setup=None):
         """
-        Dynamically create the provider info response.
+        Overriden to use the proper message class.
```
schlenk: 'Overridden'

src/oic/oauth2/provider.py
```diff
@@ -317,6 +363,145 @@ def _verify_redirect_uri(self, areq):
...
+
+        _log_info("@providerinfo_endpoint")
+        try:
+            _response = self.create_providerinfo()
+            msg = "provider_info_response: {}"
+            _log_info(msg.format(sanitize(_response.to_dict())))
+            if self.events:
+                self.events.store("Protocol response", _response)
+
+            headers = [("Cache-Control", "no-store"), ("x-ffo", "bar")]
```
schlenk: Any idea what the x-ffo header is for?

src/oic/oic/provider.py
tpazderka: Fixed.

src/oic/oauth2/provider.py
tpazderka: Hm, could stand for `X-Forwarded-For`? But the setting (and the header itself) are nonsense.

I am for removing it as it is not in a list of standard and common non-standard headers.

PR #652, opened 2019-04-15
Add bandit to the quality tools

schlenk:
Adding the bandit tool to the toolbox.
https://github.com/PyCQA/bandit

And fix some trivial py3 syntax errors in the op3 example so it doesn't blow up right away when running the bandit.

As we have some security sensitive code here, we probably should run some basic scanner like that to catch the simpler issues. Its not good enough for the CI yet, but give it a try.

Running it only on code & examples, as the tests do not need and cannot be secure.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ setup(
         'develop': ["cherrypy==3.2.4", "pyOpenSSL"],
         'testing': tests_requires,
         'docs': ['Sphinx', 'sphinx-autobuild', 'alabaster'],
-        'quality': ['pylama', 'isort', 'eradicate', 'mypy', 'black'],
+        'quality': ['pylama', 'isort', 'eradicate', 'mypy', 'black', 'bandit'],
         'ldap_authn': ['pyldap'],
     },
     install_requires=[
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ help:
 	@echo "  test       to run the tests"
 	@echo "  isort      to sort imports"
 	@echo "  blacken    to format the code"
+	@echo "  bandit     to run some simple security checkers"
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -75,7 +75,7 @@ def run_tests(self):
         'develop': ["cherrypy==3.2.4", "pyOpenSSL"],
         'testing': tests_requires,
         'docs': ['Sphinx', 'sphinx-autobuild', 'alabaster'],
-        'quality': ['pylama', 'isort', 'eradicate', 'mypy', 'black'],
+        'quality': ['pylama', 'isort', 'eradicate', 'mypy', 'black', 'bandit'],
```
tpazderka: I am not sure if this is necessary. I consider this as requirements for quality run in tox and it is not (yet) needed there.

tpazderka:
AFAIK bandit runs in Codacy, but getting it to our quality build might be a good idea.

setup.py
schlenk: Not sure either. If you have a better place to put it, I'm fine with it.

schlenk:
Ah, yes, you are right. Codacy already runs bandit on the code, those are the same issues, not sure what kind of config they use.

https://app.codacy.com/app/tpazderka/pyoidc/issues/index

tpazderka:
Codacy is a non-blocking. We can work on the issues and eventually get the code pass the bandit (or decorate with `#nosec` when it does make sense) and make the bandit a part of our quality build so we get the blocking.

setup.py
tpazderka: Maybe do not put it in requirements at all? Otherwise I do not have any other place so if you want it to be somewhere, then go ahead and merge this.

PR #653, opened 2019-04-19
Better checks

tpazderka:
---
The CI fails on one error in MyPy which should be fixed by #648.

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -7,9 +7,11 @@ from json import JSONDecodeError
 from typing import Any  # noqa - Used for MyPy
 from typing import Dict  # noqa - Used for MyPy
 from typing import List  # noqa - Used for MyPy
-from typing import Tuple
-from typing import Type
-from typing import Union
+from typing import Optional  # noqa - Used for MyPy
+from typing import Tuple  # noqa - Used for MyPy
+from typing import Type  # noqa - Used for MyPy
+from typing import Union  # noqa - Used for MyPy
+from typing import cast  # noqa - Used for MyPy
 from urllib.parse import parse_qs
 from urllib.parse import urlparse
 
@@ -921,7 +923,7 @@ class Client(oauth2.Client):
     def user_info_request(self, method="GET", state="", scope="", **kwargs):
         uir = self.message_factory.get_request_type("userinfo_endpoint")()
         logger.debug("[user_info_request]: kwargs:%s" % (sanitize(kwargs),))
-        token = None
+        token = None  # type: Optional[Token]
         if "token" in kwargs:
             if kwargs["token"]:
                 uir["access_token"] = kwargs["token"]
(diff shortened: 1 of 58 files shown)


src/oic/oic/provider.py
```diff
@@ -258,15 +262,15 @@ def __init__(
...
+        self.keyjar = self.server.keyjar  # type: KeyJar
 
         self.endp.extend([UserinfoEndpoint, RegistrationEndpoint, EndSessionEndpoint])
 
         self.userinfo = userinfo
         self.template_renderer = template_renderer
         self.baseurl = baseurl or name
-        self.cert = []
-        self.cert_encryption = []
+        self.cert = []  # type: ignore  # UNUSED?
```
schlenk: Probably unused, doesn't ring a bell here either. If those are not used in the examples, they could go.

src/oic/utils/http_util.py
```diff
@@ -225,12 +229,16 @@ def extract(environ, empty=False, err=False):
     :param empty: Stops on empty fields (default: Fault)
     :param err: Stops on errors in fields (default: Fault)
     """
+    # FIXME: This seems to be unused -> Deprecate?
```
schlenk: Thats basic wsgi parsing, so yes, deprecate if not in use.

src/oic/utils/keyio.py
```diff
@@ -451,8 +457,8 @@ def __init__(
             ``requests`` documentation.
         :return:
         """
-        self.spec2key = {}
-        self.issuer_keys = {}
+        self.spec2key = {}  # type: ignore  # FIXME: This seems unused -> Deprecate?
```
schlenk: yes, deprecate seems ok

tpazderka: Couldn't find a use so I removed it completely.

src/oic/oic/__init__.py
```diff
@@ -1447,8 +1450,8 @@ def register(self, url, registration_token=None, **kwargs):
 
         headers = {"content-type": "application/json"}
         if registration_token is not None:
-            headers["Authorization"] = b"Bearer " + b64encode(
-                registration_token.encode()
+            headers["Authorization"] = (
+                "Bearer " + b64encode(registration_token.encode()).decode()
```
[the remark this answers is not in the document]
tpazderka: Wondering myself as well :)

src/oic/oic/provider.py
tpazderka: OK, it turns out it is used in example1:
https://github.com/OpenIDC/pyoidc/blob/699cc99850d26565d3fbdc6357ca85a56bd6e5e9/oidc_example/op1/claims_provider.py#L186-L188

But it looks like it shouldn't be used in this way and rather configured statically? It is not set anywhere in the example...

src/oic/utils/http_util.py
tpazderka: Done.

src/oic/oic/provider.py
schlenk: Yes, that should just be some static resource.

schlenk:
Looks good, feel free to merge.

PR #656, opened 2019-04-25
Allow webfinger search on forced host

### LINKED ISSUE #655: Allow webfinger search on forced host
According to the specification : https://tools.ietf.org/html/rfc7033#section-4

The host is not necessary the one on the ressource

For instance, we should be able to : 

GET https://hostONE.com/.well-known/webfinger?ressource=acct:me@hostTWO.com

>  The host to which a WebFinger query is issued is significant.  If the
   query target contains a "host" portion (Section 3.2.2 of RFC 3986),
   then the host to which the WebFinger query is issued **SHOULD** be the
   same as the "host" portion of the query target, **unless the client
   receives instructions through some out-of-band mechanism to send the
   query to another host.**  If the query target does not contain a "host"
   portion, then the client chooses a host to which it directs the query
   using additional information it has.

jblemee:
---

close #655

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#631] Refactored message type handling in Client/Provider.
 
 ### Added
-- [#655] host can be force
+- [#655] Host can be forced on webfinger discovery
 - [#441] CookieDealer now accepts secure and httponly params
 
 [#655]: https://github.com/OpenIDC/pyoidc/issues/655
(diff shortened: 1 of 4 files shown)


schlenk:
Thanks for the PR. The travis errors are unrelated to your code, so nothing to fix for you there.

jblemee:
Thanks for the review, I've added tests an change a bit the changelog.

Tell me if I can do better

tpazderka:
Thanks for the update, it looks good to me.

Will merge later today.

PR #660, opened 2019-05-09
Shared sessiondb

tpazderka:
---

### DIFF
--- a/src/oic/utils/sdb.py
+++ b/src/oic/utils/sdb.py
@@ -5,6 +5,9 @@ import json
 import logging
 import time
 import uuid
+import warnings
+from abc import ABCMeta
+from abc import abstractmethod
 from typing import Dict  # noqa
 from typing import List  # noqa
 
@@ -403,7 +406,7 @@ def create_session_db(
     """
     code_factory = DefaultToken(secret, password, typ="A", lifetime=grant_expires_in)
     token_factory = DefaultToken(secret, password, typ="T", lifetime=token_expires_in)
-    db = {} if db is None else db
+    db = DictSessionBackend() if db is None else db
 
     return SessionDB(
         base_url,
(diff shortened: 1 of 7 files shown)


tpazderka:
@schlenk Could you please have a quick look? If it looks reasonable to you, I will add tests and will fox the `FIXME` things I have noticed.

src/oic/utils/sdb.py
```diff
@@ -415,6 +419,79 @@ def create_session_db(
...
+    @abstractmethod
+    def __contains__(self, key: str) -> bool:
+        """Test presence of key in storage."""
+
+    @abstractmethod
+    def get_by_uid(self, uid: str) -> List[str]:
+        """Return session ids (keys) based on `uid` (internal user identifier)."""
+
+    @abstractmethod
+    def get_by_sub(self, uid: str) -> List[str]:
```
schlenk: Parameter should be sub not uid.

schlenk:
Looks mostly reasonable as an API.

One small point that might be problematic is the fact, that the API assumes the authn event is always stored serialized as json, when iterating to find the sids for an uid. Maybe it would be better to have a two layer api, a facade that takes AuthnEvent objects and handles serialization/deserialization and the string based set/get interface for implementers to override.

tpazderka:
The AuthnEvent stuff is already handled in SessionDB directly: `create_authz_session` serializes the AuthnEvent to json, so nothing else shoudl be possible.

This is already two levels deep API (which I don't like...) and adding another layer doesn't look like a good idea to me.

schlenk:
I see your point. Agreed, we do not want even more layers there. The responsibilities are not really well factored between the TokenFactories, SessionDB and the persistence layer. Your other change to cleanup the RefreshDB/Token stuff goes in the same direction of cleaning this up.

This new API works as a better persistence option.

I think a few more of the `get_` methods taking a uid parameter and a few other methods of the current SessionDB could be moved to the new class as well, as those look a lot like DB access patterns one might like to optimize e.g. `get_client_ids_for_uid()`. Making the persistance part less dumb about what it stores.

tpazderka:
Yes, that is a good point. We should probably move that over as well.

The question is if we want to force users to implement that methods, or if we implement a dumb lookup and let them optimize it if they find the need to?

I am leaning towards the second option.

schlenk:
I think the second option (provide some dumb/clear to read default implementations) is best. Then users with a smart backend can optimize and people with a dumb backend just use it as is.

tpazderka:
OK, I have moved more stuff, I think that should be all. I will work on tests of the new code.

tpazderka:
OK, tests are finished. For some reason, Codecov is not picking up the new coverage but it should be mostly covered now, so please review.

PR #664, opened 2019-05-23
Logout message

rohe:
- [ X] Any changes relevant to users are recorded in the `CHANGELOG.md`.
- [ X] New code is annotated.
- [ X] Changes are covered by tests.
---

The messages needed to implement Single-Sign-Out according to:

- https://openid.net/specs/openid-connect-backchannel-1_0.html,
- https://openid.net/specs/openid-connect-frontchannel-1_0.html and 
- https://openid.net/specs/openid-connect-session-1_0.html

### DIFF
--- a/src/oic/oic/message.py
+++ b/src/oic/oic/message.py
@@ -38,6 +38,7 @@ from oic.oauth2.message import NotAllowedValue
 from oic.oauth2.message import ParamDefinition
 from oic.oauth2.message import SchemeError
 from oic.utils import time_util
+from oic.utils.time_util import utc_time_sans_frac
 
 __author__ = "rohe0002"
 
@@ -269,10 +270,13 @@ def check_char_set(string, allowed):
             raise NotAllowedValue("'%c' not in the allowed character set" % c)
 
 
+TOKEN_VERIFY_ARGS = ["key", "keyjar", "algs", "sender"]
+
+
 def verify_id_token(instance, check_hash=False, **kwargs):
     # Try to decode the JWT, checks the signature
     args = {}
-    for arg in ["key", "keyjar", "algs", "sender"]:
+    for arg in TOKEN_VERIFY_ARGS:
         try:
             args[arg] = kwargs[arg]
         except KeyError:
@@ -342,8 +346,9 @@ class AccessTokenResponse(message.AccessTokenResponse):
     c_param.update({"id_token": SINGLE_OPTIONAL_STRING})
 
     def verify(self, **kwargs):
-        super(AccessTokenResponse, self).verify(**kwargs)
+        super().verify(**kwargs)
         if "id_token" in self:
+            self.raw_id_token = self["id_token"]
             # replace the JWT with the verified IdToken instance
             self["id_token"] = verify_id_token(self, **kwargs)
 
@@ -367,7 +372,7 @@ class AuthorizationResponse(message.AuthorizationResponse, message.AccessTokenRe
     )
 
     def verify(self, **kwargs):
-        super(AuthorizationResponse, self).verify(**kwargs)
+        super().verify(**kwargs)
 
         if "aud" in self:
             if "client_id" in kwargs:
@@ -376,6 +381,7 @@ class AuthorizationResponse(message.AuthorizationResponse, message.AccessTokenRe
                     return False
 
         if "id_token" in self:
+            self.raw_id_token = self["id_token"]
             self["id_token"] = verify_id_token(self, check_hash=True, **kwargs)
 
         if "access_token" in self:
@@ -444,7 +450,7 @@ class AuthorizationRequest(message.AuthorizationRequest):
         All parameter values that are present both in the OAuth 2.0
         Authorization Request and in the OpenID Request Object MUST exactly match.
         """
-        super(AuthorizationRequest, self).verify(**kwargs)
+        super().verify(**kwargs)
 
         args = {}
         for arg in ["key", "keyjar", "opponent_id", "sender"]:
@@ -550,7 +556,7 @@ class OpenIDSchema(Message):
     }
 
     def verify(self, **kwargs):
-        super(OpenIDSchema, self).verify(**kwargs)
+        super().verify(**kwargs)
 
         if "birthdate" in self:
             # Either YYYY-MM-DD or just YYYY or 0000-MM-DD
@@ -604,6 +610,10 @@ class RegistrationRequest(Message):
         "initiate_login_uri": SINGLE_OPTIONAL_STRING,
         "request_uris": OPTIONAL_LIST_OF_STRINGS,
         "post_logout_redirect_uris": OPTIONAL_LIST_OF_STRINGS,
+        "frontchannel_logout_uri": SINGLE_OPTIONAL_STRING,
+        "frontchannel_logout_session_required": OPTIONAL_LOGICAL,
+        "backchannel_logout_uri": SINGLE_OPTIONAL_STRING,
+        "backchannel_logout_session_required": OPTIONAL_LOGICAL,
     }
     c_default = {"application_type": "web", "response_types": ["code"]}
     c_allowed_values = {
@@ -612,7 +622,7 @@ class RegistrationRequest(Message):
     }
 
     def verify(self, **kwargs):
-        super(RegistrationRequest, self).verify(**kwargs)
+        super().verify(**kwargs)
 
         if "initiate_login_uri" in self and not self["initiate_login_uri"].startswith(
             "https:"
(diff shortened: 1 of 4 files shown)


src/oic/oic/message.py
```diff
@@ -364,7 +370,9 @@ def verify(self, **kwargs):
                     return False
 
         if "id_token" in self:
+            _raw_id_token = self['id_token']
             self["id_token"] = verify_id_token(self, check_hash=True, **kwargs)
+            self['_raw_id_token'] = _raw_id_token
```
tpazderka: Do we really need to store it? If yes, then maybe as a class attribute so it doesn't show up in the message itself?

src/oic/oic/message.py
```diff
@@ -727,19 +749,44 @@ def verify(self, **kwargs):
...
-    c_param = {"id_token": SINGLE_REQUIRED_STRING,
-               "redirect_url": SINGLE_REQUIRED_STRING,
-               "state": SINGLE_REQUIRED_STRING}
+    c_param = {
+        "id_token": SINGLE_REQUIRED_STRING,
+        "redirect_url": SINGLE_REQUIRED_STRING,
+        "state": SINGLE_REQUIRED_STRING
+    }
+
+    def verify(self, **kwargs):
```
tpazderka: This can be probably placed into a common metaclass as it is used in multiple messages.

src/oic/oic/message.py
```diff
@@ -727,19 +749,44 @@ def verify(self, **kwargs):
...
+        super(RefreshSessionRequest, self).verify(**kwargs)
```
tpazderka: `super().verify(**kwargs)` is sufficient. This is on multiple places.

src/oic/oic/message.py
```diff
@@ -914,6 +967,119 @@ class ResourceRequest(Message):
...
+        if _keys[0] != BACK_CHANNEL_LOGOUT_EVENT:
+            raise ValueError('Wrong member in "events"')
+        if self['events'][_keys[0]] != {}:
+            raise ValueError('Wrong member value in "events"')
+
+        # There must be either a 'sub' or a 'sid', and may contain both
+        if not ('sub' in self or 'sid' in self):
+            raise ValueError('There MUST be either a "sub" or a "sid"')
+
+        try:
```
tpazderka: This `try .. except` can be replaced using `if kwargs.get("aud") is not None and kwargs["aud"] not in self["aud"]`.

Same applies in other places.

src/oic/oic/message.py
```diff
@@ -914,6 +967,119 @@ class ResourceRequest(Message):
...
+            if kwargs['iss'] != self['iss']:
+                raise NotForMe('Wrong issuer')
+        except KeyError:
+            pass
+
+        _now = utc_time_sans_frac()
+
+        try:
```
tpazderka: Can be replaced by a single line `_skew = kwargs.get("skew", 0)`, same for `iat`.

tests/test_oic_message.py
```diff
@@ -953,3 +962,112 @@ def test_verify_token_encrypted_no_key():
         verify_id_token(msg, keyjar=KeyJar(),
                         iss="https://sso.qa.7pass.ctf.prosiebensat1.com",
                         client_id="554295ce3770612820620000")
+
+
+def test_logout_token():
```
tpazderka: Please make a class for each message and add individual scenarios as test cases. That makes it much clearer which part is failing.

src/oic/oic/message.py
rohe: Yes, need to store. Since it's later going to be used as an id_token_hint.
Class attribute is a good idea.

src/oic/oic/message.py
rohe: This is nothing new that I added. But if we're changing other stuff why not this ?!

src/oic/oic/message.py
rohe: Nothing new added by me. But while cleaning up ...

tests/test_oic_message.py
rohe: OK

src/oic/oic/message.py
rohe: So, what's the benefit of the change ?

src/oic/oic/message.py
tpazderka: Hm, weird... appeared in the diff as a new code. Maybe due to conflicts?

src/oic/oic/message.py
tpazderka: The same as above... it appeared as a new code, hence the comment :)

tpazderka:
Still needs a rebase and `make isort && make blacken` :)

rohe:
make: *** No rule to make target `blacken'.  Stop.

src/oic/oic/message.py
rohe: I expected it to be a matter of taste. :-)
Myself I don't think a statement like `if kwargs.get("aud") is not None and kwargs["aud"] not in self["aud"]` is very readable.
But if you'd like me to change. I'll do so.

tpazderka:
The code is sitting on an old master and needs to be rebased to get the `blacken` target available.

src/oic/oic/message.py
tpazderka: Codacy used to complain about the `try .. except pass` statment (it seems to be fine in this instance).

If `self["aud"]` is not empty, then `if kwargs.get("aud") not in self["aud"]` should be sufficient, so it depends.

I am not going to insist on this change, though. I just thought about possible simplifications.

rohe:
I thought I had done the rebase. Hmm

tests/test_oic_message.py
```diff
@@ -953,3 +960,127 @@ def test_verify_token_encrypted_no_key():
         verify_id_token(msg, keyjar=KeyJar(),
                         iss="https://sso.qa.7pass.ctf.prosiebensat1.com",
                         client_id="554295ce3770612820620000")
+
+
+class TestLogoutToken():
```
tpazderka: The test for presence of `nonce` is missing.

tests/test_oic_message.py
```diff
@@ -953,3 +960,127 @@ def test_verify_token_encrypted_no_key():
...
+    def test_no_sub_or_sid(self):
+        lt = LogoutToken(iss='https://example.com', aud=['https://rp.example.org'],
+                         events={BACK_CHANNEL_LOGOUT_EVENT: {}},
+                         iat=utc_time_sans_frac(), jti=rndstr(16))
+
+        with pytest.raises(ValueError):
+            lt.verify()
+
+    def test_extra_event(self):
+        # more the one event
```
tpazderka: Test for wrong content of a `BACK_CHANNEL_LOGOUT_EVENT` is missing.

tests/test_oic_message.py
```diff
@@ -953,3 +960,127 @@ def test_verify_token_encrypted_no_key():
...
+        lt = LogoutToken(iss='https://example.com', aud=['https://rp.example.org'],
+                         events={BACK_CHANNEL_LOGOUT_EVENT: {}},
+                         iat=utc_time_sans_frac()+86400, jti=rndstr(16),
+                         sub='https://example.com/sub')
+
+        with pytest.raises(ValueError):
+            lt.verify()
+
+
+class TestBackchannelLogout(object):
```
tpazderka: The test for ID token verification failure is missing.

rohe: You mean Logout Token ?
There is no ID Token involved.

tpazderka: Yes, sorry. I meant LogoutToken. A test where LogoutToken fails to verify is still missing.

src/oic/oic/message.py
```diff
@@ -980,6 +1015,117 @@ class ResourceRequest(Message):
...
+    """Defines the message used in https://openid.net/specs/openid-connect-backchannel-1_0.html ."""
+
+    c_param = {"logout_token": SINGLE_REQUIRED_STRING}
+
+    def verify(self, **kwargs):
+        super().verify(**kwargs)
+
+        args = {}
+        for arg in TOKEN_VERIFY_ARGS:
+            try:
```
tpazderka: This can be rewritten using `get()` so no `try .. execept` is necessary.

rohe: Well then you must add an `if .. then ..` so what's the benefit ?

schlenk: You could use a dict comprehension:

    args = {arg: kwargs[arg] for arg in TOKEN_VERIFY_ARG if arg in kwargs}

If you really want to shorten it, but i doubt it is more readable.

tpazderka: You can update the args with `args.update(kwargs.get(arg, {})` which won't do anything in case of missing key.

Sorry, I sort of have a very strong aversion to `try .. except pass` statements and try to avoid them.

rohe: `args.update(kwargs.get(arg, {}))` fails if kwargs[arg] actually exist.
@schlenk Yes, it's not more readable. But I like comprehensions. :-)

tpazderka: Yeah, would be more complicated in this case, since you need a dict. Comprehension is fine :)

rohe:
Note that what's failing now is the integration tests. The reason might be that adjustments that has been made to this PR has not yet been reflected in oidctest. I don't know where the integration tests get there underlying libraries but there just might be a mismatch there.

tpazderka:
Everything is installed from PyPi. It is weird to have such a failure where there was no change to the code used in the integration tests. The code on master passes fine, I have restarted the build to rule out a fluke.

The only thing that is missing here is a test for `LogoutToken.verify()` that returns `False` in `BackChannelLogout`, otherwise it looks fine.

tpazderka:
The build is green now, so probably a fluke in the tests. So, please add the single missing test and we are good to go.

rohe:
Actually I don't think there is any case where `LogoutToken.verify()` will return `False`.
Made changes accordingly.

tpazderka:
Yes, looking at the code, I do agree. It either returns True or raises an exception.

I will squash and rebase the branch and merge it afterwards.

tpazderka:
Well, that took more effort than anticipated... :(

For some reason, I had to solve quite a lot of conflicts. There were some merge commits merging the branch into itself? So please try to avoid those :)

Please have a look if it seems sane, I will merge after that.

rohe:
Don't know what happened here. There where quite a lot of changes to files other than oic/message.py . But seems OK.

PR #667, opened 2019-06-14
Deprecated RefreshDB

### LINKED ISSUE #644: Deprecate refresh_db
I think it is about time to end the duality around Refresh tokens. Both access code and access token use TokenFactory, so we should go that route as well.

The only issue I can see right now, without to much digging into is the lack of session info storage. For example, this test would fail for `DefaultToken`:

https://github.com/OpenIDC/pyoidc/blob/e94b8129c26d7c9acd1ae3c4dbfe914efedc7381/tests/test_sdb.py#L272-L287

We can either implement a storage to the `DefaultToken` or we will just assume that the `SessionDB` internal database never gets cleared?

tpazderka:
Updated all code to work with refresh_token_factory

Close #644

- [ ] New code is annotated.
---

### DIFF
--- a/src/oic/utils/sdb.py
+++ b/src/oic/utils/sdb.py
@@ -8,11 +8,13 @@ import uuid
 import warnings
 from abc import ABCMeta
 from abc import abstractmethod
+from binascii import Error
 from typing import Dict  # noqa
 from typing import List  # noqa
 from typing import Optional  # noqa
 
 from cryptography.fernet import Fernet
+from cryptography.fernet import InvalidToken
 
 from oic import rndstr
 from oic.exception import ImproperlyConfigured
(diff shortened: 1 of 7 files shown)


src/oic/utils/sdb.py
```diff
@@ -148,11 +156,23 @@ def is_expired(self, token, when=None):
...
-        self.type_and_key(token)
-        return True
+        try:
+            typ, key = self.type_and_key(token)
+        except (Error, InvalidToken):
+            raise WrongTokenType()
+        if typ != self.type:
+            raise WrongTokenType()
+        if typ == "R":
+            return not self.token_storage[key].get("revoked", False)
```
schlenk: Should this care for KeyErrors, e.g. token has a key, but key isn't present in the token_storage?

tpazderka: I was thinking about the same. My original though was to have an ABC with an API and handle one specific error, but we already have several levels of ABCs...

So unless we want to add another layer of ABC for API, I think keeping this is probably better.

schlenk: Agreed.

PR #675, opened 2019-08-02
Fix reference in Changelog

tpazderka:
- [ ] New code is annotated.
---
We should probably have a look at some linter for markdown so this doesn't happen again...


schlenk:
It would probably be even better if those references could be automatically created. Should be possible if it just references the pull request itself.

But yes, basic checker is a good idea.

tpazderka:
I have added a checker (and fixed the errors it reported...). Auto-creating the references is also a good idea.

PR #680, opened 2019-08-09
Properly handle incorrect message decoding.

### LINKED ISSUE #602: Unpacking of invalid message throws uncaught exception
When an invalid JSON is passed to message init, the resulting Exception is uncaught. It should be properly handled and raises as `DecodeError`.

tpazderka:
Raise DecodeError on all failures.

Close #602

- [x] New code is annotated.
---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -1567,7 +1567,7 @@ class Provider(AProvider):
         )
         try:
             request = request_cls().deserialize(request, "json")
-        except ValueError:
+        except MessageException:
             request = request_cls().deserialize(request)
 
         logger.info("registration_request:%s" % sanitize(request.to_dict()))
--- a/tests/fakeoicsrv.py
+++ b/tests/fakeoicsrv.py
@@ -8,6 +8,7 @@ from jwkest import jws
 from jwkest.jws import alg2keytype
 
 from oic import rndstr
+from oic.oauth2.message import DecodeError
 from oic.oauth2.message import by_schema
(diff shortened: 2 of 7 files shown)


src/oic/oic/provider.py
```diff
@@ -1567,7 +1568,7 @@ def create_registration(self, authn=None, request=None, **kwargs):
         )
         try:
             request = request_cls().deserialize(request, "json")
-        except ValueError:
+        except DecodeError:
```
schlenk: from_dict() may raise ValueError if values in the dict are unexpected (e.g. bool values that are not bool). Those errors would now raise here. So maybe catch both.

tests/fakeoicsrv.py
```diff
@@ -215,7 +216,7 @@ def userinfo_endpoint(self, data):
     def registration_endpoint(self, data):
         try:
             req = self.parse_registration_request(data, "json")
-        except ValueError:
+        except DecodeError:
```
schlenk: as above, might raise ValueErrors as well for malformed content after sucessful json decoding.

tpazderka:
Hm, interesting question. I think that it would be better to throw a single type of exception (or base class). I will go over the code and adjust that.

tests/fakeoicsrv.py
tpazderka: I am not going to fix this as I want to drop it. See #682

tpazderka:
Updated to raise a (hopefully) single base exception.

PR #682, opened 2019-08-16
Remove fakeserver

tpazderka:
- [ ] New code is annotated.
---
Replace it by testing with responses.

It tends to break on changes to provider and is just basically a separate implementation.


src/oic/oauth2/__init__.py
```diff
@@ -839,6 +839,7 @@ def do_access_token_request(
             http_args = ht_args
         else:
             http_args.update(ht_args)
+            http_args.pop("password", None)
```
schlenk: why pop password here?

tpazderka: For some reason it was passing through and ended up as an invalid kwarg to requests library...

schlenk: ok. It is wrong in any case, so the fix should do no harm, but probably the issue is burried somewhere in request_info().

tpazderka: Yes, my guess as well. But didn't want to spend a huge amount of time on that. Will make a separate issue for that...

PR #688, opened 2019-09-19
Logout consumer 2

rohe:
- [x] New code is annotated.
---

This is step 2 of adding single sign out to pyoidc. This part includes changes to session databases used by OPs and RPs and necessary changes to the OIDC Consumer class.

### DIFF
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -654,7 +654,8 @@ class Client(PBase):
             if "id_token" in resp and self.sso_db:
                 session_update(self.sso_db, _state, "sub", resp["id_token"]["sub"])
                 session_update(self.sso_db, _state, "issuer", resp["id_token"]["iss"])
-
+                if "sid" in resp["id_token"]:
+                    session_update(self.sso_db, _state, "smid", resp["id_token"]["sid"])
         return resp
 
     def init_authentication_method(
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -49,6 +49,7 @@ from oic.utils.http_util import BadRequest
 from oic.utils.http_util import Response
 from oic.utils.http_util import SeeOther
 from oic.utils.keyio import KeyJar
+from oic.utils.session_backend import session_update
 from oic.utils.time_util import utc_time_sans_frac
 
 __author__ = "rohe0002"
@@ -195,6 +196,7 @@ class Client(PBase):
             timeout=timeout,
         )
 
+        self.sso_db = None
         self.client_id = client_id
         self.client_authn_method = client_authn_method
 
@@ -649,8 +651,9 @@ class Client(PBase):
             except KeyError:
                 self.grant[_state] = self.grant_class(resp=resp)
 
-            if 'id_token' in resp:
-                self.sdb.update(_state, "sub", resp["id_token"]["sub"])
+            if "id_token" in resp:
+                session_update(self.sso_db, _state, "sub", resp["id_token"]["sub"])
+                session_update(self.sso_db, _state, "issuer", resp["id_token"]["iss"])
 
         return resp
 
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -651,7 +651,7 @@ class Client(PBase):
             except KeyError:
                 self.grant[_state] = self.grant_class(resp=resp)
 
-            if "id_token" in resp:
+            if "id_token" in resp and self.sso_db:
                 session_update(self.sso_db, _state, "sub", resp["id_token"]["sub"])
                 session_update(self.sso_db, _state, "issuer", resp["id_token"]["iss"])
 
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -49,7 +49,7 @@ from oic.utils.http_util import BadRequest
 from oic.utils.http_util import Response
 from oic.utils.http_util import SeeOther
 from oic.utils.keyio import KeyJar
-from oic.utils.session_backend import session_update
+from oic.utils.sdb import session_update
 from oic.utils.time_util import utc_time_sans_frac
 
 __author__ = "rohe0002"
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -26,9 +26,9 @@ from oic.utils import http_util
 from oic.utils.sanitize import sanitize
 from oic.utils.sdb import DictSessionBackend
 from oic.utils.sdb import SessionBackend
-from oic.utils.session_backend import session_extended_get
-from oic.utils.session_backend import session_get
-from oic.utils.session_backend import session_update
+from oic.utils.sdb import session_extended_get
+from oic.utils.sdb import session_get
+from oic.utils.sdb import session_update
 
 __author__ = "rohe0002"
 
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -548,6 +548,5 @@ class Consumer(Client):
             _sid = session_extended_get(
                 self.sso_db, sub, "issuer", req["logout_token"]["iss"]
             )
-            # _sid = self.sdb.get_by_sub_and_(sub, "issuer", req["logout_token"]["iss"])
 
         return _sid
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -2,8 +2,6 @@ import logging
 import os.path
 import warnings
 
-from jwkest import as_unicode
-
 from oic import rndstr
 from oic.exception import AuthzError
 from oic.exception import MessageException
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -1,5 +1,6 @@
 import logging
 import os.path
+import warnings
 
 from jwkest import as_unicode
 
@@ -16,13 +17,18 @@ from oic.oauth2.message import ErrorResponse
 from oic.oic import ENDPOINTS
 from oic.oic import Client
 from oic.oic.message import AccessTokenResponse
-from oic.oic.message import BackChannelLogoutRequest
 from oic.oic.message import AuthorizationRequest
 from oic.oic.message import AuthorizationResponse
+from oic.oic.message import BackChannelLogoutRequest
 from oic.oic.message import Claims
 from oic.oic.message import ClaimsRequest
 from oic.utils import http_util
 from oic.utils.sanitize import sanitize
+from oic.utils.sdb import DictSessionBackend
+from oic.utils.sdb import SessionBackend
+from oic.utils.session_backend import session_extended_get
+from oic.utils.session_backend import session_get
+from oic.utils.session_backend import session_update
 
 __author__ = "rohe0002"
 
@@ -82,7 +88,16 @@ def clean_response(aresp):
     return atr
 
 
-IGNORE = ["request2endpoint", "response2error", "grant_class", "token_class"]
+IGNORE = [
+    "request2endpoint",
+    "response2error",
+    "grant_class",
+    "token_class",
+    "sdb",
+    "wf",
+    "events",
+    "message_factory",
+]
 
 CONSUMER_PREF_ARGS = [
     "token_endpoint_auth_method",
@@ -116,6 +131,7 @@ class Consumer(Client):
         server_info=None,
         debug=False,
         client_prefs=None,
+        sso_db=None,
     ):
         """
         Initialize a Consumer instance.
@@ -146,7 +162,23 @@ class Consumer(Client):
                 except KeyError:
                     setattr(self, endpoint, "")
 
+        if not isinstance(session_db, SessionBackend):
+            warnings.warn(
+                "Please use `SessionBackend` to ensure proper API for the database.",
+                DeprecationWarning,
+            )
         self.sdb = session_db
+
+        if sso_db:
+            if not isinstance(sso_db, SessionBackend):
+                warnings.warn(
+                    "Please use `SessionBackend` to ensure proper API for the database.",
+                    DeprecationWarning,
+                )
+            self.sso_db = sso_db
+        else:
+            self.sso_db = DictSessionBackend()
+
         self.debug = debug
         self.seed = ""
         self.nonce = ""
@@ -174,6 +206,7 @@ class Consumer(Client):
             if not _val and val:
                 setattr(self, key, val)
             elif key == "grant" and val:
+                # val is a Grant instance
                 val.update(_val)
                 setattr(self, key, val)
 
@@ -238,6 +271,7 @@ class Consumer(Client):
 
         self._backup(sid)
         self.sdb["seed:%s" % self.seed] = sid
+        self.sso_db[sid] = {}
 
         args = {
             "client_id": self.client_id,
@@ -372,7 +406,7 @@ class Consumer(Client):
                 idt = None
             else:
                 try:
-                    self.sdb.update(idt['sid'], "smid", _state)
+                    session_update(self.sdb, idt["sid"], "smid", _state)
                 except KeyError:
                     pass
 
@@ -400,7 +434,7 @@ class Consumer(Client):
                 idt = None
             else:
                 try:
-                    self.sdb.update(idt['sid'], "smid", _state)
+                    session_update(self.sso_db, idt["sid"], "smid", _state)
                 except KeyError:
                     pass
 
@@ -482,7 +516,7 @@ class Consumer(Client):
     def end_session(self):
         pass
 
-# LOGOUT related
+    # LOGOUT related
 
     def backchannel_logout(self, request=None, request_args=None):
         """
@@ -498,29 +532,28 @@ class Consumer(Client):
         else:
             req = BackChannelLogoutRequest(**request_args)
 
-        kwargs = {
-            'aud': self.client_id,
-            'iss': self.issuer,
-            'keyjar': self.keyjar
-        }
+        kwargs = {"aud": self.client_id, "iss": self.issuer, "keyjar": self.keyjar}
 
         try:
             req.verify(**kwargs)
         except (MessageException, ValueError, NotForMe) as err:
-            raise MessageException('Bogus logout request: {}'.format(err))
+            raise MessageException("Bogus logout request: {}".format(err))
 
         # Find the subject through 'sid' or 'sub'
 
         try:
-            sub = req['logout_token']['sub']
+            sub = req["logout_token"]["sub"]
         except KeyError:
             try:
-                sm_id = req['logout_token']['sid']
+                sm_id = req["logout_token"]["sid"]
             except KeyError:
                 raise MessageException('Neither "sid" nor "sub"')
             else:
-                _sid = self.sdb.get_by_smid(sm_id)
+                _sid = session_get(self.sso_db, "smid", sm_id)
         else:
-            _sid = self.sdb.get_by_sub_and_x(sub, "issuer", req["logout_token"]["iss"])
+            _sid = session_extended_get(
+                self.sso_db, sub, "issuer", req["logout_token"]["iss"]
+            )
+            # _sid = self.sdb.get_by_sub_and_(sub, "issuer", req["logout_token"]["iss"])
 
         return _sid
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -167,7 +167,7 @@ class Consumer(Client):
             )
         self.sdb = session_db
 
-        if sso_db:
+        if sso_db is not None:
             if not isinstance(sso_db, SessionBackend):
                 warnings.warn(
                     "Please use `SessionBackend` to ensure proper API for the database.",
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -541,12 +541,9 @@ class Consumer(Client):
         try:
             sub = req["logout_token"]["sub"]
         except KeyError:
-            try:
-                sm_id = req["logout_token"]["sid"]
-            except KeyError:
-                raise MessageException('Neither "sid" nor "sub"')
-            else:
-                _sid = session_get(self.sso_db, "smid", sm_id)
+            # verify has guaranteed that there will be a sid if sub is missing
+            sm_id = req["logout_token"]["sid"]
+            _sid = session_get(self.sso_db, "smid", sm_id)
         else:
             _sid = session_extended_get(
                 self.sso_db, sub, "issuer", req["logout_token"]["iss"]
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -528,7 +528,7 @@ class Consumer(Client):
         :return: A Session Identifier
         """
         if request:
-            req = BackChannelLogoutRequest().from_urlencoded(as_unicode(request))
+            req = BackChannelLogoutRequest().from_urlencoded(request)
         else:
             req = BackChannelLogoutRequest(**request_args)
 
--- a/src/oic/oic/consumer.py
+++ b/src/oic/oic/consumer.py
@@ -434,7 +434,7 @@ class Consumer(Client):
                 idt = None
             else:
                 try:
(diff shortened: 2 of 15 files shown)


tpazderka:
The quality builds are failing and need to be fixed.

src/oic/utils/sdb.py
```diff
@@ -553,6 +553,57 @@ def get_by_uid(self, uid: str) -> List[str]:
             if AuthnEvent.from_json(session["authn_event"]).uid == uid
         ]
 
+    def get_uid_by_sub(self, sub):
```
tpazderka: I am not sure these are really needed as we already have methods to return the whole session, so this is just a little helper method. Depends on number of uses.

But if these are really needed/beneficial, they need to be moved to the base class to ensure proper API and should use the already existing `get_by_sub` and `get_by_uid` methods to benefit from optimizations.

src/oic/utils/sdb.py
```diff
@@ -1018,44 +1077,50 @@ def get_client_id_for_session(self, sid):
 
     def get_client_ids_for_uid(self, uid: str) -> List[str]:
         """Return client_ids for a given uid."""
-        return self._db.get_client_ids_for_uid(uid)
```
tpazderka: These changes seem unnecessary. They were already implemented using the SessionDB API so why reimplement them differently?

src/oic/oic/consumer.py
```diff
@@ -461,3 +481,46 @@ def check_session(self):
...
+        """
+        Receives a back channel logout request and returns a Session ID
+        if the request was OK.
+
+        :param request: A urlencoded request
+        :param request_args: The request as a dictionary
+        :return: A Session Identifier
+        """
+        if request:
+            req = BackChannelLogoutRequest().from_urlencoded(as_unicode(request))
```
tpazderka: The `as_unicode` is not necessary. `from_urlencoded` can handle both.

src/oic/utils/sdb.py
```diff
@@ -1082,6 +1147,8 @@ def duplicate(self, sinfo):
                 pass
 
         self._db[sid] = _dic
+        self._db.update_sub2sid_map(_dic["sub"], sid)
```
tpazderka: This method doesn't seem to exist.

src/oic/utils/sdb.py
```diff
@@ -1091,3 +1158,28 @@ def read(self, token):
...
+    def get_by_sub(self, sub):
+        return self._db.get_by_sub(sub)
+
+    def get_by_uid(self, uid):
+        return self._db.get_by_uid(uid)
+
+    def get_by_smid(self, smid):
+        return self._db.get_by_smid(smid)
+
+    def get_by_sub_and_client_id(self, sid, client_id):
```
tpazderka: These are sort of a retrieve and verify and should probably be in a client and not in SessionDB.

src/oic/utils/sdb.py
```diff
@@ -1091,3 +1158,28 @@ def read(self, token):
...
+
+    def get_by_sub_and_issuer(self, sid, issuer):
+        return self._db.get_by_sub_and_x(sid, "issuer", issuer)
+
+    def make_smid(self, sid):
+        """
+        Create a session management ID
+
+        :param sid:
+        :return: A session management ID
```
tpazderka: This doesn't seem to be used (yet) is there a usage in the second part?

src/oic/utils/session_backend.py
```diff
@@ -0,0 +1,293 @@
...
+        else:
+            item = self.storage[key]
+            item[attribute] = value
+            self.storage[key] = item
+
+
+def session_update(db, key, attr, val):
+    if isinstance(db, SessionBackend):
+        db.update(key, attr, val)
+    elif isinstance(db, dict):
```
tpazderka: The `else` statements should probably also raise `DeprecationWarning` so we can remove them later.

src/oic/utils/sdb.py
rohe: Yes

src/oic/utils/sdb.py
rohe: OK

src/oic/utils/sdb.py
rohe: Correct

src/oic/oic/consumer.py
rohe: OK

src/oic/utils/sdb.py
rohe: They are as you say helper method and I think useful because they are used by other parts of the code.
Always prefer to implement in one place if possible.

src/oic/utils/sdb.py
rohe: I would argue that the changes make the implementation more robust.
Handles a possible exception that the original doesn't.

tpazderka: Raising `KeyError` for missing `uid` seems fine for me and we should always have `client_id` in session. So the benefit doesn't look that great to me.

But if you really want to do this, than it should be done in `SessionBackend`.

src/oic/oic/consumer.py
```diff
@@ -461,3 +515,45 @@ def check_session(self):
...
+
+        try:
+            sub = req["logout_token"]["sub"]
+        except KeyError:
+            try:
+                sm_id = req["logout_token"]["sid"]
+            except KeyError:
+                raise MessageException('Neither "sid" nor "sub"')
+            else:
+                _sid = session_get(self.sso_db, "smid", sm_id)
```
tpazderka: Since this is in a separate database, we can probably still call it `sid` instead of `smid`. This would reduce the number of methods that need to be implemented.

And in reality you probably really want two separate storages, so there is little risk of people targeting single storage with both actions.

rohe: True

rohe: But I'd keep smidges since sid used for the key to the database. Would be confusing if sid was both the key and a property in the DB.
Session ID (sid) and Session Management ID (smid) are different things.

src/oic/utils/session_backend.py
rohe: OK

rohe:
pyjama reports a number of "E0100 SyntaxError" none of which seems to have anything to do with additions in this PR.

rohe:
The oidctest tests fails because the oidctest code assumes that the logout support for both RP and OP exist in PyOIDC. And that is not the case here. You miss the provider part.

rohe:
I've checked in a change to oidctest that should take care of the oidctest running problem.

tpazderka:
The quality build is broken on master as well. I have it fixed, just waiting for a review and then I will merge it.

You are missing some tests as indicated by the Codecov report.

src/oic/oauth2/__init__.py
```diff
@@ -649,6 +651,10 @@ def parse_response(
             except KeyError:
                 self.grant[_state] = self.grant_class(resp=resp)
 
+            if "id_token" in resp:
```
tpazderka: This probably needs an additional check: `if self.sso_db is not None and "id_token" in resp:`, otherwise it can fail for OAuth?

src/oic/utils/session_backend.py
```diff
@@ -0,0 +1,201 @@
...
+            if AuthnEvent.from_json(session["authn_event"]).uid == uid
+        ]
+
+    def get(self, attr: str, val: str) -> List[str]:
+        """Return session ids based on attribute name and value."""
+        return [
+            sid for sid, session in self.storage.items() if session.get(attr) == val
+        ]
+
+    def get_uid_by_sub(self, sub):
```
tpazderka: These methods need to be moved to the `SessionBackend`, otherwise the `SessionDB` wouldn't work with other implementations. These methods should also use `__getitem__` or other methods instead of access to `storage` which is implementation specific.

Also they should be annotated like the rest of the code here.

tests/test_sdb.py
```diff
@@ -171,70 +171,66 @@ def test_expired_when(self):
         assert factory.is_expired(token, when=when) is True
 
 
-class TestSessionBackend(TestCase):
```
tpazderka: I don't think any of the changes here are necessary?

The only thing needed here is addition of tests for the newly implemented methods.

src/oic/oic/consumer.py
tpazderka: OK, we can keep it that way.

src/oic/oauth2/__init__.py
rohe: You're right.

src/oic/utils/session_backend.py
rohe: OK

rohe:
Sorry for spelling error. issuers should be issues.

src/oic/utils/session_backend.py
```diff
@@ -0,0 +1,208 @@
...
+        :param value: Attribute value
+        """
+        if key not in self:
+            self[key] = {attribute: value}
+        else:
+            item = self[key]
+            item[attribute] = value
+            self[key] = item
+
+    def get_uid_by_sub(self, sub):
```
tpazderka: I think this can be implemented here using `get_by_sub` same as below. There shouldn't be a need for backend specific implementation.

src/oic/utils/session_backend.py
```diff
@@ -0,0 +1,208 @@
...
+    def get_token_ids(self, uid: str) -> List[str]:
+        """Return id_tokens for the given uid."""
+        return [self[sid]["id_token"] for sid in self.get_by_uid(uid)]
+
+    def is_revoke_uid(self, uid: str) -> bool:
+        """Return if the session is revoked."""
+        # We do not care which session it is - once revoked, al are revoked
+        return any([self[sid]["revoked"] for sid in self.get_by_uid(uid)])
+
+    def update(self, key, attribute, value):
```
tpazderka: Missing annotation. I presume something like `update(self, key: str, attribute: str, value: Any) -> NoReturn:`

rohe: OK

src/oic/utils/session_backend.py
rohe: Sorry, I don't understand.
What exactly do you think can be implemented differently ?
In SessionBackend there is no way you can loop through the items or values which is exactly what you need to do in this method. Or have I missed something ?

tpazderka: There is already an abstractmethod `get_by_sub` that returns the whole session info, from there, you can get `uid` directly.

The benefit (from my point of view) is that users don't have to implement this always, only if they feel they can get a speed benefit.

tests/test_oic_consumer_logout.py
```diff
@@ -271,9 +269,11 @@ def test_logout_without_sub(self):
 
     def test_logout_with_none(self):
         # Now for the backchannel logout. This happens on the OP
-        # Pgitython version 3.5 doesn't support variable annotation
+        # Python version 3.5 doesn't support variable annotation
```
tpazderka: You can still use comment annotation as in the rest of the code:
`logout_info = {}  # typing: Dict[str, ...]`

tpazderka:
Also, there are still missing tests, coverage of the diff is only 70%.

src/oic/utils/session_backend.py
rohe: OK I get it, but `get_by_sub` doesn't return the session info it returns sids.
But using the sids you can get the session info's.
So it has to be indirect.

rohe:
Where  can I see the test coverage of the diff ?

tpazderka:
> Where can I see the test coverage of the diff ?

https://codecov.io/gh/OpenIDC/pyoidc/pull/688/diff

tpazderka:
Please don't use merge to update the branch. It makes mess when merging to master. It is better to rebase.

src/oic/oic/provider.py
```diff
@@ -2027,3 +2027,24 @@ def remove_inactive_keys(self, more_then=3600):
                         kb.remove(key)
             if len(kb) == 0:
                 self.keyjar.issuer_keys[""].remove(kb)
+
+    def get_uid_by_sub(self, sub):
```
tpazderka: These methods are (now) already implemented in SessionBackend, so we should use those directly.

`self.sdb.get_uid_by_sub()` whenever this method is used.

rohe: OK

tests/test_oic_consumer_logout.py
rohe: Noop! If I do then pylama complains.

rohe:
So, I can't make AppVeyor, pylama and mypy happy at the same time.
Which means one or more of them will always fail.

rohe:
Open for suggestions.

tpazderka:
The last one seemed to work :) Will go over it soon.

src/oic/oic/consumer.py
```diff
@@ -461,3 +513,40 @@ def check_session(self):
 
     def end_session(self):
         pass
+
+    # LOGOUT related
+
+    def backchannel_logout(self, request=None, request_args=None):
```
tpazderka: As a new method, this should be annotated. You can then probably drop the second part of the docstring.

src/oic/oic/provider.py
```diff
@@ -2027,3 +2027,12 @@ def remove_inactive_keys(self, more_then=3600):
                         kb.remove(key)
             if len(kb) == 0:
                 self.keyjar.issuer_keys[""].remove(kb)
+
+    def get_by_sub_and_(self, sub, key, val):
+        for sid in self.sdb.get_by_sub(sub):
```
tpazderka: This should be annotated, also needs a docstring.

The `KeyError` and `return None` are not covered by tests.

src/oic/oic/consumer.py
```diff
@@ -384,6 +430,12 @@ def parse_authz(self, query="", **kwargs):
                 idt = aresp["id_token"]
             except KeyError:
                 idt = None
+            else:
+                try:
+                    session_update(self.sso_db, _state, "smid", idt["sid"])
+                except KeyError:
```
tpazderka: Is this really necessary? If yes, than it needs tests.

src/oic/utils/session_backend.py
```diff
@@ -0,0 +1,190 @@
...
+        :param value: Attribute value
+        """
+        if key not in self:
+            self[key] = {attribute: value}
+        else:
+            item = self[key]
+            item[attribute] = value
+            self[key] = item
+
+    def get_uid_by_sub(self, sub):
```
tpazderka: The new methods need annotation.

src/oic/utils/session_backend.py
```diff
@@ -0,0 +1,190 @@
...
+        """Return User ids based on sub. Should only be one."""
+        uid = ""
+        for sid in self.get_by_sub(sub):
+            if uid:
+                if uid != AuthnEvent.from_json(self[sid]["authn_event"]).uid:
+                    raise ValueError("More the one uid bound to one sub")
```
tpazderka: This raise is not tested. Can this really happen in real life?

src/oic/utils/sdb.py
```diff
@@ -613,6 +474,7 @@ def __init__(
 
         self.access_token = self.token_factory["access_token"]
         self.token = self.access_token
+        self.sm_salt = sm_salt or rndstr(32)
```
tpazderka: This probably should be set first and then set the `sm_salt` on `_db`?

src/oic/utils/sdb.py
```diff
@@ -1018,44 +884,49 @@ def get_client_id_for_session(self, sid):
...
 
-    def get_verify_logout(self, uid: str) -> str:
-        """Return the key for logout verification."""
-        return self._db.get_verified_logout(uid)
+    def get_verify_logout(self, uid: str) -> Dict:
+        """Return the dictionary for logout verification."""
+        res = {}
+        for sid in self._db.get_by_uid(uid):
+            _dict = self._db[sid]
+            try:
```
tpazderka: This can be rewritten using `get()`.

The except part is not tested.

src/oic/utils/sdb.py
```diff
@@ -1018,44 +884,49 @@ def get_client_id_for_session(self, sid):
...
+            _dict = self._db[sid]
+            _dict["verified_logout"] = uuid.uuid4().urn
+
+    def get_id_token(self, uid: str) -> Dict:
+        res = {}
+        for sid in self._db.get_by_uid(uid):
+            _dict = self._db[sid]
+            try:
+                res[_dict["client_id"]] = _dict["id_token"]
+            except KeyError:
```
tpazderka: Test for missing `id_token` is missing.

src/oic/utils/sdb.py
```diff
@@ -1018,44 +884,49 @@ def get_client_id_for_session(self, sid):
...
+                pass
+        return res
 
-    def is_revoke_uid(self, uid: str) -> bool:
-        """Return if the uid session has been revoked."""
-        return self._db.is_revoke_uid(uid)
+    def is_revoke_uid(self, uid: str) -> Dict:
+        res = {}
+        for sid in self._db.get_by_uid(uid):
+            try:
```
tpazderka: This can be rewritten using `get()`.

The except part is not tested.

src/oic/utils/sdb.py
```diff
@@ -1091,3 +964,117 @@ def read(self, token):
             raise WrongTokenType("Not a grant token")
 
         return self._db[key]
+
+    def get_by_sub(self, sub):
```
tpazderka: These new methods need annotation and docstrings.

src/oic/utils/sdb.py
```diff
@@ -1091,3 +964,117 @@ def read(self, token):
...
+        return self._db.get_by_uid(uid)
+
+    def get_uid_by_sub(self, sub):
+        return self._db.get_uid_by_sub(sub)
+
+    def get_uid_by_sid(self, sub):
+        return self._db.get_uid_by_sid(sub)
+
+
+def session_get(db, attr, val):
```
tpazderka: It looks like these are not used that much. There is probably quite a lot of calls to the old API?

src/oic/oic/consumer.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

src/oic/oic/consumer.py
rohe: Done

src/oic/utils/session_backend.py
rohe: Unlikely, so we could leave it out until we've seen it happen ?

src/oic/utils/sdb.py
rohe: Ah, yes!

src/oic/utils/sdb.py
rohe: This is a provider issue. Not part of my PR.
So let's get back to it when we do the next part.

src/oic/utils/sdb.py
rohe: Another provider related issue

src/oic/utils/sdb.py
rohe: And a third provider related issue.

src/oic/utils/sdb.py
rohe: OK

src/oic/utils/sdb.py
rohe: Might become more used when we deal with the provider side.

src/oic/utils/sdb.py
tpazderka: If it is not part of PR, then it shouldn't be here. Unfortunately these things rarely get revisited... :)

src/oic/utils/sdb.py
tpazderka: Could be, we will see. Maybe we will need to do a followup cleanup to properly use them everywhere.

src/oic/utils/session_backend.py
tpazderka: I guess... I am for removing this piece of code.

src/oic/oic/provider.py
```diff
@@ -568,7 +570,7 @@ def end_session_endpoint(self, request="", cookie=None, **kwargs):
             "endsession_endpoint"
         )().from_urlencoded(request)
 
-        logger.debug("End session request: {}", sanitize(esr.to_dict()))
+        logger.debug("End session request: {}".format(sanitize(esr.to_dict())))
```
tpazderka: Logging should be lazy, no need to evaluate the expression if we are not going to log it.

rohe: Changed it because one of the mypy or pylama, don't remember which one, complained about too many arguments.

tpazderka: Just checked and it doesn't complain for me. Might be from an earlier iteration of the code?

rohe: Might be. I'm not wed to one or the other version so I'll change back if that is what you want.

tpazderka: Hm, looking at it, it was wrong previously as well. Probably a reason for the complaining.

It should be:
`logger.debug("End session request: %s", sanitize(esr.to_dict()))`

It is not possible to use `{}` in logging...

rohe: Which is why I removed the `{}` in the latest commit.

tpazderka: Yes, but you are missing the `%s` which will cause a `TypeError`. Also the `format()` is redundant.

Either do not change this at all and we will deal with it in separate issue, or add the `%s` to the message string.

rohe: I'll add `%s` to the message string.

src/oic/utils/sdb.py
tpazderka: Docstring doesn't match the original implementation.

src/oic/utils/sdb.py
tpazderka: The same applies here: If not part of PR, shouldn't be here.

src/oic/utils/sdb.py
rohe: Well, it matches the underlying method and is correct.

src/oic/utils/sdb.py
tpazderka: This is still missing.

src/oic/utils/sdb.py
rohe: OK

src/oic/utils/sdb.py
tpazderka: I am talking about `get_verify_logout`, which has a return of `str` and the docstring claims that it returns `dictionary`.

src/oic/utils/sdb.py
rohe: OK

src/oic/utils/sdb.py
rohe: OK

src/oic/utils/sdb.py
rohe: OK

src/oic/utils/sdb.py
rohe: I guess this is outdated.

src/oic/utils/session_backend.py
rohe: OK

src/oic/utils/sdb.py
tpazderka: Hm, I still see it as changed...

src/oic/utils/sdb.py
tpazderka: This is still here.

rohe: Removed

rohe:
I don't get what mypy is complaining about now that is different from a while ago.
If I run mypy locally I get: Success !!

tpazderka:
New release of mypy :(

We will have to deal with this on master and then rebase this on top.

Should be able to do it tonight or tomorrow.

rohe:
So are you fine with the PR or ?

tpazderka:
Yes, you can consider this one finished. It is very unlikely something will pop up after I fix the quality on master.

You can start your second PR on top of this one.

rohe:
I'll wait until you've done the merging and rebasing.

tpazderka:
Merged by hand.

Issue #692, opened 2019-10-07
If birthdate is `None`, validation fails


timothy-spencer:
In `src/oic/oic/message.py`, the OpenIDSchema verify function seems to try to verify that the birthdate is a proper date format, but this fails if it is `None`.  I'm not sure why I'm running into this, but not others, but it is preventing me from claiming my auth token from a login.gov IDP.

I'm generating a PR right now for this which seems to fix it, but I am not 100% sure that I understand what is going on here, like why this isn't something that everybody runs into, so I will definitely be looking for feedback from folks who know the code better!

Thanks!

schlenk:
The spec clearly defines the birthdate claim as a string:

> 

birthdate | string | End-User's birthday, represented as an 	  ISO 8601:2004 [ISO8601‑2004] YYYY-MM-DD 	  format. The year MAY be 0000, indicating that it is omitted. 	  To represent only the year, YYYY format is allowed. Note that 	  depending on the underlying platform's date related function, providing just year can 	  result in varying month and day, so the implementers need to take this factor into account 	  to correctly process the dates.
-- | -- | --

(see https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims )

So ``None`` is not a valid value for this claim and the validation should fail.

tpazderka:
AFAIK if there is no value, the claim should not be included.

timothy-spencer:
Hmm.  Interesting.  I thought it was just checking the format of the claim.  That seems reasonable, though it is unsatisfying to me, because I'd really love to just get this working.  :-)  I guess I'll try to dig into this some more to see why birthdate is in there and is None.  Thanks!

selfissued:
@tpazderka is correct.  As Connect Core says at https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse, "If a Claim is not returned, that Claim Name SHOULD be omitted from the JSON object representing the Claims; it SHOULD NOT be present with a null or empty string value."

Allan-Nava:
> AFAIK if there is no value, the claim should not be included.

if the value is None how can I fix this problem?

tpazderka:
The claim should not be included. That needs to be fixed on the other side.

Allan-Nava:
> The claim should not be included. That needs to be fixed on the other side.

So inside the keycloak admin console?

tpazderka:
Not sure what can and cannot be done in the console, but I would suspect that it is an error in their implementation.

[...]

PR #693, opened 2019-10-07
Fix validation of birthdate to handle case when it is 'None'

timothy-spencer:
- [X] Any changes relevant to users are recorded in the `CHANGELOG.md`.
- [ ] New code is annotated.
---

This is to address https://github.com/OpenIDC/pyoidc/issues/692


timothy-spencer:
It was pointed out that the spec says that this is supposed to validate the entire claim, not just it's format, so this is actually proper behavior, and I need to figure out why birthdate is None.  Sigh.

PR #698, opened 2019-10-18
'state' in kwargs MUST NOT overwrite 'state' in request_args.

rohe:
'state' in kwargs is used as a key to access information in the RP.
'state' in request_args is just a request argument.
They are not necessarily the same.

- [ ] New code is annotated.
---

### DIFF
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -628,7 +628,7 @@ class Client(oauth2.Client):
         if request_args is None:
             request_args = {}
 
-        if "state" in request_args and 'state' not in kwargs:
+        if "state" in request_args and "state" not in kwargs:
             kwargs["state"] = request_args["state"]
 
         return self._id_token_based(request, request_args, extra_args, **kwargs)
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -628,9 +628,7 @@ class Client(oauth2.Client):
         if request_args is None:
             request_args = {}
 
-        if "state" in kwargs:
-            request_args["state"] = kwargs["state"]
-        elif "state" in request_args:
+        if "state" in request_args and 'state' not in kwargs:
             kwargs["state"] = request_args["state"]
 
         return self._id_token_based(request, request_args, extra_args, **kwargs)
--- a/tests/test_oic.py
+++ b/tests/test_oic.py
@@ -621,7 +621,7 @@ class TestClient(object):
             ],
         )
 
-    def test_construct_EndSessionRequest(self):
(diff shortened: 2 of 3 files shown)


tests/test_oic.py
```diff
@@ -632,8 +632,20 @@ def test_construct_EndSessionRequest(self):
 
         self.client.grant["foo"].tokens.append(Token(resp))
 
+        # state only in kwargs
         args = {"redirect_url": "http://example.com/end"}
         esr = self.client.construct_EndSessionRequest(state="foo", request_args=args)
+        assert _eq(esr.keys(), ["id_token", "redirect_url"])
```
tpazderka: Please make these as a separate tests under the same class. It is then much cleaner to figure out what fails.

src/oic/oic/__init__.py
```diff
@@ -628,9 +628,7 @@ def construct_EndSessionRequest(
         if request_args is None:
             request_args = {}
 
-        if "state" in kwargs:
-            request_args["state"] = kwargs["state"]
-        elif "state" in request_args:
+        if "state" in request_args and "state" not in kwargs:
```
tpazderka: This is also achievable by using `setdefault("state", kwargs.get("state"))` and you can go without the condition completely.

Althought that would set `state  = None` if it would be missing in both.

But I am not sure if that is readable enough, so leaving that up to you.

tests/test_oic.py
rohe: OK

tpazderka: Also shouldn't this include `state` as well?

src/oic/oic/__init__.py
rohe: I'd go for what I have proposed since I think it's more readable.

tpazderka: Fine with me.

tests/test_oic.py
rohe: No, `state` in kwargs are different from `state` in request_args.
This comes from the fact that we have state in different requests being different things.
`state` in an authorisation request is NOT the same thing as `state` in an end session requests.
The `state` that comes in the authorisation request are used internally as a key to information resources so the construct_EndSessionRequest needs that `state`. The `state` in the request_args is specific to the end session request and is not expected to have the same value as the authorisation request state.

rohe: Hence, `state` in kwargs should not bleed over into the request args.

rohe:
The errors in travis-ci is not of my doing.

tests/test_oic.py
tpazderka: OK.

PR #700, opened 2019-10-20
The provider side of handling logout

rohe:
- [x] New code is annotated.
---

### DIFF
--- a/src/oic/oic/provider.py
+++ b/src/oic/oic/provider.py
@@ -4,11 +4,15 @@ import json
 import logging
 import socket
 import time
+import uuid
 from functools import cmp_to_key
+from http.cookies import SimpleCookie
 from typing import Any
 from typing import Dict  # noqa
 from typing import List  # noqa
 from typing import Optional
+from typing import Tuple
+from typing import Union
 from urllib.parse import parse_qs
 from urllib.parse import splitquery  # type: ignore
 from urllib.parse import unquote
@@ -34,11 +38,11 @@ from oic.exception import InvalidRequest
 from oic.exception import MessageException
 from oic.exception import NotForMe
 from oic.exception import ParameterError
-from oic.exception import SubMismatch
 from oic.exception import UnSupported
 from oic.oauth2 import compact
 from oic.oauth2 import error_response
 from oic.oauth2 import redirect_authz_error
+from oic.oauth2.base import PBase
 from oic.oauth2.exception import CapabilitiesMisMatch
 from oic.oauth2.exception import VerificationError
 from oic.oauth2.message import Message
@@ -52,6 +56,7 @@ from oic.oic import PROVIDER_DEFAULT
 from oic.oic import Server
 from oic.oic import claims_match
 from oic.oic import scope2claims
+from oic.oic.message import BACK_CHANNEL_LOGOUT_EVENT
 from oic.oic.message import SCOPE2CLAIMS
 from oic.oic.message import AccessTokenResponse
 from oic.oic.message import AuthorizationResponse
@@ -69,6 +74,7 @@ from oic.utils.http_util import Created
 from oic.utils.http_util import Response
 from oic.utils.http_util import SeeOther
 from oic.utils.http_util import Unauthorized
+from oic.utils.jwt import JWT
 from oic.utils.keyio import KEYS  # noqa
 from oic.utils.keyio import KeyBundle
 from oic.utils.keyio import KeyJar  # noqa
@@ -238,6 +244,9 @@ class Provider(AProvider):
         template_renderer=render_template,
         extra_scope_dict=None,
         message_factory=OIDCMessageFactory,
+        post_logout_page="",
+        self_signing_alg="RS256",
+        logout_path="",
     ):
 
         # This has to be defined before calling super()
@@ -293,6 +302,13 @@ class Provider(AProvider):
         # Allow custom schema (inheriting from OpenIDSchema) to be used -
         # additional attributes
         self.schema = schema
+
+        # Logout connected attributes
+        self.httpc = PBase(verify_ssl=verify_ssl, keyjar=self.keyjar)
+        self.post_logout_page = post_logout_page
+        self.signing_alg = self_signing_alg
+        self.logout_path = logout_path
+        self.logout_verify_url = ""
 
     @property
     def default_capabilities(self):
@@ -488,194 +504,11 @@ class Provider(AProvider):
 
         return req_user
 
-    def verify_post_logout_redirect_uri(self, esreq, client_id):
-        """
-        Verify the post_logout_redirect_uris.
-
-        :param esreq: End session request
-        :param client_id: The Client ID
-        :return:
-        """
-        try:
-            redirect_uri = esreq["post_logout_redirect_uri"]
-        except KeyError:
-            logger.debug("Missing post_logout_redirect_uri parameter")
-            return
-
-        try:
-            accepted_urls = self.cdb[client_id]["post_logout_redirect_uris"]
-            if self._verify_url(redirect_uri, accepted_urls):
-                return redirect_uri
-        except Exception as exc:
-            msg = "An error occurred while verifying redirect URI: %s"
-            logger.debug(msg, str(exc))
-
-        return None
-
     def is_session_revoked(self, request="", cookie=None):
         areq = parse_qs(request)
         authn, _ = self.pick_auth(areq)
         identity, _ts = authn.authenticated_as(cookie)
         return self.sdb.is_revoke_uid(identity["uid"])
-
-    def let_user_verify_logout(self, uid, esr, cookie, redirect_uri):
-        if cookie:
-            headers = [cookie]
-        else:
-            headers = []
-
-        self.sdb.set_verified_logout(uid)
-
-        if redirect_uri is not None:
-            redirect = redirect_uri
-        else:
-            redirect = "/"
-        try:
-            tmp_id_token_hint = esr["id_token_hint"]
-        except KeyError:
-            tmp_id_token_hint = ""
-
-        context = {
-            "id_token_hint": tmp_id_token_hint,
-            "post_logout_redirect_uri": esr["post_logout_redirect_uri"],
-            "key": self.sdb.get_verify_logout(uid),
-            "redirect": redirect,
-            "action": "/" + EndSessionEndpoint("").etype,
-        }
-        return Response(
-            self.template_renderer("verify_logout", context), headers=headers
-        )
-
-    def _get_sids_from_cookie(self, cookie):
-        """Get cookie_dealer, client_id and sids from cookie."""
-        if cookie is None:
-            return None, None, None
-
-        cookie_dealer = CookieDealer(srv=self)
-        client_id = sids = None
-
-        _cval = cookie_dealer.get_cookie_value(cookie, self.sso_cookie_name)
-        if _cval:
-            (value, _ts, typ) = _cval
-            if typ == "sso":
-                uid, client_id = value.split(DELIM)
-                try:
-                    sids = session_get(self.sdb, "uid", uid)
-                except (KeyError, IndexError):
-                    raise SubMismatch("Mismatch uid")
-        return cookie_dealer, client_id, sids
-
-    def end_session_endpoint(self, request="", cookie=None, **kwargs):
-        esr = self.server.message_factory.get_request_type(
-            "endsession_endpoint"
-        )().from_urlencoded(request)
-
-        logger.debug("End session request: %s", format(sanitize(esr.to_dict())))
-
-        # 2 ways of find out client ID and user. Either through a cookie
-        # or using the id_token_hint
-        try:
-            cookie_dealer, client_id, sids = self._get_sids_from_cookie(cookie)
-        except SubMismatch as error:
-            return error_response("invalid_request", "%s" % error)
-
-        if "id_token_hint" in esr:
-            id_token_hint = IdToken().from_jwt(
-                esr["id_token_hint"], keyjar=self.keyjar, verify=True
-            )
-            far_away = 86400 * 30  # 30 days
-
-            if client_id:
-                args = {"client_id": client_id}
-            else:
-                args = {}
-
-            try:
-                id_token_hint.verify(
-                    iss=self.baseurl, skew=far_away, nonce_storage_time=far_away, **args
-                )
-            except (VerificationError, NotForMe) as err:
-                logger.warning("Verification error on id_token_hint: {}".format(err))
-                return error_response("invalid_request", "Bad Id Token hint")
-
-            sub = id_token_hint["sub"]
-
-            if sids:
-                match = False
-                # verify that 'sub' are bound to 'user'
-                for sid in sids:
-                    if self.sdb[sid]["sub"] == sub:
-                        match = True
-                        break
-                if not match:
-                    return error_response("invalid_request", "Wrong user")
-            else:
-                try:
-                    sids = self.sdb.get_by_sub(sub)
-                except IndexError:
-                    pass
-
-            if not client_id:
-                if len(id_token_hint["aud"]) == 1:
-                    client_id = id_token_hint["aud"][0]
-                else:
-                    client_id = id_token_hint["azp"]
-
-        if not client_id:
-            return error_response("invalid_request", "Could not find client ID")
-        if client_id not in self.cdb:
-            return error_response("invalid_request", "Unknown client")
-
-        match = False
-        for sid in sids:
-            if self.sdb[sid]["client_id"] == client_id:
-                match = True
-                break
-        if not match:
-            return error_response("invalid_request", "Unmatched client")
-
-        redirect_uri = None
-        if "post_logout_redirect_uri" in esr:
-            redirect_uri = self.verify_post_logout_redirect_uri(esr, client_id)
-            if not redirect_uri:
-                msg = "Post logout redirect URI verification failed!"
-                return error_response("invalid_request", msg)
-        else:  # If only one registered use that one
-            if len(self.cdb[client_id]["post_logout_redirect_uris"]) == 1:
-                _base, _query = self.cdb[client_id]["post_logout_redirect_uris"][0]
-                if _query:
-                    query_string = urlencode(
-                        [(key, v) for key in _query for v in _query[key]]
-                    )
-                    redirect_uri = "%s?%s" % (_base, query_string)
-                else:
-                    redirect_uri = _base
-
-        for sid in sids:
-            del self.sdb[sid]
-
-        # Delete cookies
-        authn, _ = self.pick_auth(esr)
-        headers = [authn.delete_cookie(), self.delete_session_cookie()]
-        if cookie_dealer:
-            headers.append(cookie_dealer.delete_cookie(self.sso_cookie_name))
-
-        if redirect_uri is not None:
-            try:
-                _state = esr["state"]
-            except KeyError:
-                redirect_uri = str(redirect_uri)
-            else:
-                if "?" in redirect_uri:
-                    redirect_uri += "&"
-                else:
-                    redirect_uri += "?"
-
-                redirect_uri += urlencode({"state": _state})
-
-            return SeeOther(redirect_uri, headers=headers)
-
-        return Response("Successful logout", headers=headers)
 
     def verify_endpoint(self, request="", cookie=None, **kwargs):
         """
@@ -852,7 +685,7 @@ class Provider(AProvider):
         aresp["client_id"] = areq["client_id"]
 
         if self.events:
-            self.events.store("Protocol response", aresp)
+            self.events.store("protocol response", aresp)
 
         response = sanitize(aresp.to_dict())
         logger.info("authorization response: %s", response)
@@ -2018,7 +1851,7 @@ class Provider(AProvider):
             # print to the jwks file
             dump_jwks(self.keyjar[""], self.jwks_name)
 
-    def remove_inactive_keys(self, more_then=3600):
+    def remove_inactive_keys(self, more_then: int = 3600):
         """
         Remove all keys that has been inactive 'more_then' seconds.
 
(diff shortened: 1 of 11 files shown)


rohe:
The remaining Codacy error is the log line we discussed earlier.
You want me to change it ?

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+            "key": self.sdb.get_verify_logout(uid),
+            "redirect": redirect,
+            "action": "/" + EndSessionEndpoint("").etype,
+        }
+        return Response(
+            self.template_renderer("verify_logout", context), headers=headers
+        )
+
+    def _get_uid_from_cookie(
+        self, cookie: Union[str, SimpleCookie, None]
```
tpazderka: This is probably `Optional[Union[str, SimpleCookie]]`?

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+            if _iframe:
+                res["flu"] = {client_id: _iframe}
+
+        # Clean out sessions
+        self.clean_sessions([sid])
+
+        return res
+
+    def end_session_endpoint(
+        self, request: str = "", cookie: Union[str, SimpleCookie, None] = None, **kwargs
```
tpazderka: Again, isn't this `Optional[Union[str, SimpleCookie]]`?

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+        """
+        Handle a RP initiated Logout request.
+
+        :param request: The logout request
+        :param cookie:
+        :param kwargs:
+        :return: Returns a dictionary with one key 'sjwt' and the value
+            being a signed JWT token with session information.
+        """
+        esr = EndSessionRequest().from_urlencoded(request)
```
tpazderka: Please use `self.server.message_factory.get_request_type("endsession_endpoint")` for easier overrides.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                args = {"client_id": client_id}
+            else:
+                args = {}
+
+            try:
+                id_token_hint.verify(
+                    iss=self.baseurl, skew=far_away, nonce_storage_time=far_away, **args
+                )
+            except (VerificationError, NotForMe) as err:
+                logger.warning("Verification error on id_token_hint: {}".format(err))
```
tpazderka: Yes, please fix this so it actually logs properly.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                return error_response("invalid_request", "Bad Id Token hint")
+
+            sub = id_token_hint["sub"]
+
+            if uid:
```
tpazderka: Test for `None` should be explicit.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                else:
+                    client_id = id_token_hint["azp"]
+
+            sids = session_get(self.sdb, "sub", sub)
+
+            for sid in sids:
+                if self.sdb[sid]["client_id"] == client_id:
+                    break
+
+        if not client_id:
```
tpazderka: Again, test for `None` should be explicit.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                msg = "Post logout redirect URI verification failed!"
+                return error_response("invalid_request", msg)
+        else:  # If only one registered use that one
+            try:
+                _ruri = self.cdb[client_id]["post_logout_redirect_uris"]
+            except KeyError:
+                msg = "Missing post_logout_redirect_uri"
+                return error_response("invalid_request", msg)
+
+            if len(_ruri) == 1:
```
tpazderka: `redirect_uri` is not defined for case where the `len(_ruri)` is different from `1`.

This is also not covered in tests. Either there are no logout URIs registered (and the SDB doesn't raise KeyError, which might be possible) or there are multiple.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+
+        return back_channel_logout_uri, sjwt
+
+    def clean_sessions(self, usids: List[str]):
+        """
+        Remove Session IDs from the session DB.
+
+        :param usids: List of session IDs
+        """
+        _sdb = self.sdb
```
tpazderka: This assignment seem unnecessary...

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+
+        :param sid: Session ID
+        :return: Dictionary with logout information
+        """
+        _sdb = self.sdb
+
+        uid = _sdb.get_uid_by_sid(sid)
+        _client_sid = {}
+        # Find all the session IDs this user has gotten
+        usids = session_get(_sdb, "uid", uid)
```
tpazderka: I think we do not need to use `sessoin_get` for provider, since using anything else than proper SDB is discouraged for two releases.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+        }  # type: Dict[str, Dict[str, Union[None, str, Tuple[str,str]]]]
+
+        if "backchannel_logout_uri" in _cdb[client_id]:
+            _sub = _sso_db[sid]["sub"]
+            _spec = self.do_back_channel_logout(_cdb[client_id], _sub, sid)
+            if _spec:
+                res["blu"] = {client_id: _spec}
+        elif "frontchannel_logout_uri" in _cdb[client_id]:
+            # Construct an IFrame
+            _iframe = do_front_channel_logout_iframe(_cdb[client_id], self.name, sid)
```
tpazderka: I would rather have this as a staticmethod of provider, keeps it together.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                _iframe = do_front_channel_logout_iframe(_cdb[_cid], _iss, _csid)
+                if _iframe:
+                    fc_iframes[_cid] = _iframe
+
+        # Clean out all sessions
+        self.clean_sessions(usids)
+
+        return {"blu": bc_logouts, "flu": fc_iframes}
+
+    def logout_from_client(self, sid: str, client_id: str) -> Dict:
```
tpazderka: This method would benefit for more descriptive variable names... It is sort of unreadable...

src/oic/oic/provider.py
```diff
@@ -2049,3 +1883,449 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+
+    def do_verified_logout(
+        self, sid: str, client_id: str, alla: bool = False, **kwargs
+    ) -> Dict:
+        """
+        Perform back channel logout and prepares the information needed for front channel logout.
+
+        :param sid: Session ID
+        :param client_id:
+        :param alla:
```
tpazderka: Have no idea what this stands for... Either make it more descritptive, or add the docstring.

tests/test_oic_provider_logout.py
```diff
@@ -0,0 +1,831 @@
...
+AUTHN_BROKER = AuthnBroker()
+AUTHN_BROKER.add("UNDEFINED", DummyAuthn(None, "username"))
+
+# dealing with authorization
+AUTHZ = AuthzHandling()
+SYMKEY = rndstr(16)  # symmetric key used to encrypt cookie info
+USERINFO = UserInfo(USERDB)
+
+
+class DummyHTTPClient(object):
```
tpazderka: This class can be easily replaced using `responses` which allow for cleaner mocking of HTTP requests without the need for modification of the objects. I have got rid of almost all of them and I don't want to add more in new tests.

tests/test_oic_provider_logout.py
```diff
@@ -0,0 +1,831 @@
...
+    def __init__(self, resp_status=200):
+        self.resp_status = resp_status
+
+    def http_request(self, *args, **kwargs):
+        return Response(status_code=self.resp_status)
+
+
+class TestProvider(object):
```
tpazderka: I try to avoid using pytest in favor of unittest. The are (in my opinion) better and provide handy assert functions so you don't have to write `assert` statements.

I won't enforce this, but please consider it.

src/oic/oic/provider.py
rohe: Sort of the same but yes. Optional is just a shorthand.

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: It makes things go a bit faster but if you don't care about that.

src/oic/oic/provider.py
rohe: You previously marked this as a breaking change and therefor something that should be avoided. Has that changed ?
It was the whole reason for me to do all session_* functions.

tests/test_oic_provider_logout.py
rohe: Haven't used unittest these last 6-7 years. Switched to py.test because it was so much better. That might not be the case anymore ?

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
tpazderka: That was a reference to the Codacy issue. This needs to be changed to percent formating and lazy syntax.

src/oic/oic/provider.py
tpazderka: That would probably depend on implementation, wouldn't it?

It is probably fine either way, was just curious.

tests/test_oic_provider_logout.py
tpazderka: I believe they got much better and they are in standard library so that is a plus :0

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

tests/test_oic_provider_logout.py
rohe: Done

src/oic/oic/provider.py
rohe: Done

src/oic/oic/provider.py
tpazderka: Hm, not sure...

src/oic/oic/provider.py
```diff
@@ -2191,6 +2197,10 @@ def end_session_endpoint(
                     redirect_uri = "%s?%s" % (_base, query_string)
                 else:
                     redirect_uri = _base
+            else:
+                raise ValueError(
+                    "Missing post_logout_redirect_uri and more then one post_logout_redirect_uris"
```
tpazderka: Shouldn't this result in `error_response` back to the client?

tests/test_oic_provider_logout.py
```diff
@@ -779,9 +804,9 @@ def test_do_verified_logout_just_the_one(self):
         # I know the session backend DB is a DictSessionBackend so I can use that
         _sid = list(self.provider.sdb._db.storage.keys())[0]
 
-        self.provider.httpc = DummyHTTPClient(200)
-
+        # There is no back channel logout, hence no HTTP POST and no need for responses
```
tpazderka: Then it might be a good idea to add responses anyway as they will make sure that no communication is done. This makes this behaviour fixed as the test would otherwise fail.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+        One of uid and sid MUST be provided. If uid is provided sid is ignored.
+        NO changes are made to the session DB.
+        No logout is actually performed
+        :param uid: User ID
+        :param sid: Session ID
+        :return: Dictionary with logout information
+        """
+        if not uid:
+            if not sid:
+                raise ParameterError("One of uid and sid MUST be provided")
```
tpazderka: This part is never covered in tests.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+        # Front-/Backchannel logout ?
+        _cdb = self.cdb
+        _iss = self.name
+        bc_logouts = {}
+        fc_iframes = {}
+        for _cid, _csid in _client_sid.items():
+            if "backchannel_logout_uri" in _cdb[_cid]:
+                _sub = self.sdb[_csid]["sub"]
+                _spec = self.do_back_channel_logout(_cdb[_cid], _sub, _csid)
+                if _spec:
```
tpazderka: This is missing tests for `False` case, where `do_back_channel_logout` returns empty `_spec`. Also it should probably be an explicit test for None.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                    bc_logouts[_cid] = _spec
+            elif "frontchannel_logout_uri" in _cdb[_cid]:
```
tpazderka: Test for  missing `back` and `front` logout uri is missing.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                # Construct an IFrame
+                _iframe = self.do_front_channel_logout_iframe(_cdb[_cid], _iss, _csid)
+                if _iframe:
```
tpazderka: This is missing tests for False case, where do_front_channel_logout returns empty _iframe. Also it should probably be an explicit test for None.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+            sids = session_get(self.sdb, "sub", sub)
+
+            for sid in sids:
+                if self.sdb[sid]["client_id"] == client_id:
+                    break
+
+        if not client_id:
+            return error_response("invalid_request", "Could not find client ID")
+        if client_id not in self.cdb:
+            return error_response("invalid_request", "Unknown client")
```
tpazderka: This is not covered in tests.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                if self.sdb.get_uid_by_sub(sub) != uid:
+                    return error_response("invalid_request", "Wrong user")
+            else:
+                uid = self.sdb.get_uid_by_sub(sub)
+
+            if client_id is None:
+                if len(id_token_hint["aud"]) == 1:
+                    client_id = id_token_hint["aud"][0]
+                else:
+                    client_id = id_token_hint["azp"]
```
tpazderka: This branch is not covered in tests.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+
+            sids = session_get(self.sdb, "sub", sub)
+
+            for sid in sids:
```
tpazderka: What happens if no match is found? Also that case is not covered in tests.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+            logout_spec = self.logout_info_for_one_client(
+                session_id=sid, client_id=client_id
+            )
+            sids = [sid]
+
+        if self.events:
+            self.events.store("object args", "{}".format(logout_spec))
+
+        if not logout_spec["back_channel"] and not logout_spec["front_channel"]:
+            return {}
```
tpazderka: This is missing a test.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+            failed = []
+            for _cid, spec in logout_spec["back_channel"].items():
+                _url, sjwt = spec
+                logger.info("logging out from {} at {}".format(_cid, _url))
+
+                try:
+                    res = self.httpc.http_request(
+                        _url, "POST", data="logout_token={}".format(sjwt)
+                    )
+                except Exception as err:
```
tpazderka: This is not covered in tests.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                elif res.status_code >= 400:
+                    _errstr = "failed to logout from {}".format(_cid)
+                    if self.events:
+                        self.events.store("fault", _errstr)
+                    logger.error(_errstr)
+                    failed.append(_cid)
+            # If no back-channel logout worked and there is no front-channel logout
+            # regard this as a failure.
+            if len(failed) == len(logout_spec["back_channel"]):
+                if not logout_spec["front_channel"]:
```
tpazderka: The empty else is not covered in tests.

src/oic/oic/provider.py
```diff
@@ -2049,3 +1882,471 @@ def get_by_sub_and_(self, sub: str, key: str, val: Any) -> Optional[str]:
...
+                    # Can't be more specific because I don't know which http client are used
+                    logger.error("failed to logout from {}".format(_cid))
+                    if self.events:
+                        self.events.store("exception", "{}: {}".format(_cid, str(err)))
+                    failed.append(_cid)
+                    continue
+
+                if res.status_code < 300:
+                    logger.info("Logged out from {}".format(_cid))
+                elif res.status_code >= 400:
```
tpazderka: It looks to me that this can be a simple `else` since redirects should be followed, aren't they?

src/oic/oic/provider.py
rohe: Probably

tests/test_oic_provider_logout.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: Well, do_back_channel_logout() is only called from the logout_for_* methods if there is a backchannel_logout_uri. So the test is not needed at all.

src/oic/oic/provider.py
rohe: Same here. Tests not needed.

src/oic/oic/provider.py
tpazderka: Then the `_spec` should never be `None` and the `if _spec` clause is not needed and should be removed.

src/oic/oic/provider.py
tpazderka: Same as above, when the condition can never be False, it shouldn't be here.

src/oic/oic/provider.py
tpazderka: No, this method can be called on logins to clients with no logout mechanics registered and such case should be covered in tests.

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: OK

src/oic/oic/provider.py
rohe: Elaborate please !

tpazderka: The case when this condition is not met is never tested.

src/oic/oic/provider.py
rohe: To assume that the client automatically follows a 30X is probably fair.

tests/test_oic_provider_logout.py
```diff
@@ -0,0 +1,978 @@
...
+        self.provider.cdb["client0"]["client_id"] = "client0"
+
+        # Get a session ID, anyone will do.
+        # I know the session backend DB is a DictSessionBackend so I can use that
+        _sid = list(self.provider.sdb._db.storage.keys())[0]
+
+        # There is no back channel logout, hence there should be no HTTP POST
+        exception = ConnectionError()
+        with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
+            rsps.add(responses.POST, "https://example.com/bc_logout", body=exception)
```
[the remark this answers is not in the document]
rohe: OK

src/oic/oic/provider.py
rohe: It's been removed

tests/test_oic_provider_logout.py
```diff
@@ -0,0 +1,977 @@
...
+        "verified": True,
+        "sub": "username",
+        "extra_claim": "extra_claim_value",
+    },
+}
+
+URLMAP = {CLIENT_ID: ["https://example.com/authz"]}
+
+
+def dummy_template_renderer(template_name: str, context: Dict[str, str]) -> str:
```
tpazderka: This should probably not be here. The default template rendered can be used. And if any modifications for proper logout functionality are needed, they should be implemented (and documented) there.

tests/test_oic_provider_logout.py
```diff
@@ -0,0 +1,1012 @@
...
+            "backchannel_logout_uri"
+        ] = "https://example.com/bc_logout"
+        self.provider.cdb["client0"]["client_id"] = "client0"
+
+        # Get a session ID, anyone will do.
+        # I know the session backend DB is a DictSessionBackend so I can use that
+        _sid = list(self.provider.sdb._db.storage.keys())[0]
+
+        # There is no back channel logout, hence there should be no HTTP POST
+        exception = ConnectionError()
```
tpazderka: The exception is now unused, but quality will probably complain about that as well :)

tests/test_oic_provider_logout.py
```diff
@@ -0,0 +1,1012 @@
...
+            if c in self.provider.cdb["number5"]:
+                del self.provider.cdb["number5"][c]
+
+        resp = self.provider.do_verified_logout(
+            sid=list(self.provider.sdb._db.storage.keys())[0], client_id="number5"
+        )
+
+        assert resp == {}
+
+    def test_back_channel_http_exception(self):
```
tpazderka: The test name is misleading, it should probably be `test_no_back_channel`

src/oic/oic/provider.py
tpazderka: This test for `logout_info_for_all_clients` with one of the clients missing both logout specs is still missing.

rohe: OK

src/oic/oic/provider.py
rohe: Done

tests/test_oic_provider_logout.py
rohe: Removed

tests/test_oic_provider_logout.py
rohe: I propose test_back_channel_logout_fails

tpazderka: Sounds good.

tests/test_oic_provider_logout.py
rohe: OK

tests/test_oic_provider_logout.py
```diff
@@ -0,0 +1,1043 @@
...
+        ] = "https://example.com/bc_logout"
+        self.provider.cdb["client0"]["client_id"] = "client0"
+
+        # Get a session ID, anyone will do.
+        # I know the session backend DB is a DictSessionBackend so I can use that
+        _sid = list(self.provider.sdb._db.storage.keys())[0]
+
+        # There is no back channel logout, hence there should be no HTTP POST
+        with responses.RequestsMock() as rsps:
+            rsps.calls  # This is just to quite pylama
```
tpazderka: The simple `with responses.RequestsMock():` shoudl do the trick as well?

rohe: No, I tried that and got complains.

tpazderka: You need to remove the assignment of the decorator, then it works (just tested it).

tpazderka:
OK, I will squash it a little bit and merge.

PR #708, opened 2019-11-01
Wants the original non-parsed JWT and not an IDToken instance.

rohe:
- [ ] New code is annotated.
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on the [KeepAChangeLog] project.
 
 ## Unreleased
 
+### Fixed
+- [#708] Wants the original non-parsed JWT and not an IDToken instance.
+
+[#708]: https://github.com/OpenIDC/pyoidc/pull/708
+
 ## 1.1.0 [2019-10-25]
 
 ### Changed
--- a/tests/test_oic.py
+++ b/tests/test_oic.py
@@ -14,6 +14,7 @@ from requests import Response
 
 from oic.exception import RegistrationError
 from oic.oauth2.exception import OtherError
+from oic.oauth2.message import SINGLE_OPTIONAL_STRING
 from oic.oauth2.message import MessageTuple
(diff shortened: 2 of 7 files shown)


rohe:
I think mypy is wrong !

tpazderka:
Yes, it cannot recognize the dynamically created attributes of the Client. You can probably add the ignore statement on the offending line (`type: ignore`).

I will go through the rest of the PR later today.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on the [KeepAChangeLog] project.
 - [#679] Make `state` optional in `EndSessionRequest`
 - [#683] Fix basic_auth with client password
 - [#698] `state` in `EndSessionRequest` request args and kwargs different
+- [#708] Wants the original non-parsed JWT and not an IDToken instance.
```
tpazderka: This needs to go to the unreleased section as it is going to be a bugfix.

tests/test_oic.py
```diff
@@ -182,19 +181,14 @@ def test_access_token_request(self):
         assert _eq(resp.keys(), ["token_type", "state", "access_token", "scope"])
 
     def test_access_token_request_with_custom_response_class(self):
```
tpazderka: Please leave this test unchanged. This is here to ensure that the custom response class is honored and is not actually related to the content of the response.

rohe: Done

CHANGELOG.md
rohe: Done

tests/test_oic.py
```diff
@@ -224,7 +223,6 @@ class CustomMessageFactory(OIDCMessageFactory):
                     "access_token": "token",
                     "scope": "openid",
                     "state": "state0",
-                    "raw_id_token": "token",
```
tpazderka: This line should also remain here.

rohe: Damn, right you are !

PR #709, opened 2019-11-04
Fix readme

tpazderka:
- [ ] New code is annotated.
---
`command` is not supported on PyPi, so it won't upload...

We might want to add a check as well...


tpazderka:
Yeah, I haven't realized either... I figured it out only after I got yelled at by PyPi :)

PR #712, opened 2019-11-13
Set Content-Type for backchannel logout POSTs

rohe:
- [ ] New code is annotated.
---

According to spec the POST body uses the application/x-www-form-urlencoded encoding.


tpazderka:
Failing CI is because of #714, will be dealt with there...

PR #719, opened 2019-11-21
Allow OpenID Client registration with JWT

### LINKED ISSUE #718: Client Registration with Inital Access Token does not work with Keycloak
Hi,

The functionality to registers one's client via an initial access token
does not work with the keycloak server, maybe other Providers as well. The functionality (introduced with 7377a7a4e5e185363e)
takes an arbitrary string, base64 encodes it and adds it to a header.
Creating a registration token with keycloak already gives us an base64 encoded token. So the token gets double b64 encoded
and therefore rejected.
b64-Decoding it beforehand does not work because then we have a bytes object with a RSA Signature which cannot really be be decoded to a string object.
So, I see two possibilities:
* Add a other parameter for already b64 encoded strings
* Check if the token is b64 encoded, use it if it is, else b64- encode it.

I could implement both (or other solutions), just tell me what's more appropriate.

JustKiddingCode:
When register an OpenID Client,
registration tokens that are valid JWTs are now passed through to the
OpenID Provider. Every other string is still base64 encoded.
This closes #718 

Don't know if the previous two are applicable.
- [x] New code is annotated.
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on the [KeepAChangeLog] project.
 
 ## Unreleased
 
+### Added
+- [#719] Add support for JWT registration tokens
+
 ### Fixed
 - [#711] Deal with no post_logout_redirect_uri
 - [#712] Set Content-Type on BackChannel logout POST.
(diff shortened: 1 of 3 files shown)


tpazderka:
Thank you for the contribution!

PR #733, opened 2019-12-23
Fix Oauth flow for client with Client Credentials

### LINKED ISSUE #727: OAuth 2.0 client requesting token using client_credentials grant type
According to [1], one should be allowed to make access token request with a OAuth2 client using client_credentials flow. On the client side one should not expect any grant - client_id and client_secret should be the "grant". This does not seem to be currently the case in the pyoidc library - the oauth.Client specific part. 

When one tried to ask for access token using the OAuth2.0 client, an exception is thrown informing the user about missing grant. When trying to use CCAccessTokenRequest, it does not convert it to the endpoint because line (71) of the init file is commented out - see [2] (happens also when using client.do_access_token_request function).

Workaround [3] should be possible, although not naturally reflecting the use-cases of the code design. 

This issue is an outcome of a user question in issue #724 

[1] https://tools.ietf.org/html/rfc6749#section-4.4

> 4.4.1. Authorization Request and Response
> 
> Since the client authentication is used as the authorization grant,
> no additional authorization request is needed.
> 

[2]

```
client.do_any(CCAccessTokenRequest, request_args=args, authn_method="client_secret_basic")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 777, in do_any
    endpoint=endpoint)
  File "/usr/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 468, in request_info
    **kwargs)
  File "/usr/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 422, in uri_and_body
    uri = self._endpoint(self.request2endpoint[reqmsg.__name__],
KeyError: 'CCAccessTokenRequest'
from oic.oauth2.message import AccessTokenRequest
client.do_any(AccessTokenRequest, request_args=args, authn_method="client_secret_basic")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 777, in do_any
    endpoint=endpoint)
  File "/usr/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 442, in request_info
    **kwargs)
  File "/usr/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 362, in construct_AccessTokenRequest
    grant = self.get_grant(**kwargs)
  File "/usr/lib/python2.7/site-packages/oic/oauth2/__init__.py", line 281, in get_grant
    raise GrantError("No grant found for state:'%s'" % state)
oic.oauth2.exception.GrantError: No grant found for state:''
```

[3]
```
from oic.oauth2.message import CCAccessTokenRequest
from oic.oauth2 import REQUEST2ENDPOINT
REQUEST2ENDPOINT['CCAccessTokenRequest']='token_endpoint'
client.do_any(CCAccessTokenRequest, request_args=args, authn_method="client_secret_basic")
```

tpazderka:
Close #727

- [x] New code is annotated.
---

### DIFF
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -31,6 +31,7 @@ from oic.oauth2.message import ASConfigurationResponse
 from oic.oauth2.message import AuthorizationErrorResponse
 from oic.oauth2.message import AuthorizationRequest
 from oic.oauth2.message import AuthorizationResponse
+from oic.oauth2.message import CCAccessTokenRequest
 from oic.oauth2.message import ErrorResponse
 from oic.oauth2.message import GrantExpired
 from oic.oauth2.message import Message
@@ -68,7 +69,7 @@ REQUEST2ENDPOINT = {
     "AuthorizationRequest": "authorization_endpoint",
     "AccessTokenRequest": "token_endpoint",
     # ROPCAccessTokenRequest: "authorization_endpoint",
-    # CCAccessTokenRequest: "authorization_endpoint",
+    "CCAccessTokenRequest": "authorization_endpoint",
     "RefreshAccessTokenRequest": "token_endpoint",
     "TokenRevocationRequest": "token_endpoint",
 }
@@ -400,7 +401,11 @@ class Client(PBase):
 
     def construct_AccessTokenRequest(
         self,
-        request: Type[AccessTokenRequest] = None,
+        request: Union[
+            Type[AccessTokenRequest],
+            Type[ROPCAccessTokenRequest],
+            Type[CCAccessTokenRequest],
+        ] = None,
         request_args=None,
         extra_args=None,
         **kwargs
@@ -410,7 +415,7 @@ class Client(PBase):
             request = self.message_factory.get_request_type("token_endpoint")
         if request_args is None:
             request_args = {}
-        if request is not ROPCAccessTokenRequest:
+        if request not in (ROPCAccessTokenRequest, CCAccessTokenRequest):
             grant = self.get_grant(**kwargs)
 
             if not grant.is_valid():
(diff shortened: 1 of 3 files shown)


src/oic/oauth2/__init__.py
```diff
@@ -68,7 +69,7 @@
     "AuthorizationRequest": "authorization_endpoint",
     "AccessTokenRequest": "token_endpoint",
     # ROPCAccessTokenRequest: "authorization_endpoint",
```
schlenk: should the ROPAccessTokenRequest be commented out here, when it is used later?

src/oic/oauth2/__init__.py
```diff
@@ -68,7 +69,7 @@
...
-    # CCAccessTokenRequest: "authorization_endpoint",
+    "CCAccessTokenRequest": "authorization_endpoint",
```
schlenk: This is the wrong endpoint, at least according to the RFC 6749 4.4.2
"(A)  The client authenticates with the authorization server and
        requests an access token from the **token endpoint**."

src/oic/oauth2/__init__.py
```diff
@@ -410,7 +415,7 @@ def construct_AccessTokenRequest(
             request = self.message_factory.get_request_type("token_endpoint")
         if request_args is None:
             request_args = {}
-        if request is not ROPCAccessTokenRequest:
+        if request not in (ROPCAccessTokenRequest, CCAccessTokenRequest):
```
schlenk: If we follow the Type[x] annotation style above, this should allow for subclasses as well.

tpazderka:
Fixed all the commented issues.

I think there is any need to block this for Oic client, it should be possible to use it against Oauth providers as well.

PR #739, opened 2020-01-23
Handle HTTP 405 (method not allowed error)

JustKiddingCode:
Hi,

I wrote a client with that supports multiple providers. One of the providers restricts the HTTP Method
to GET. I don't want to restrict every provider to only use get, maybe the next will only support POST.
Since clients must return allowed methods, I looked into the http header and retry the userinfo request. If this is wanted behaviour, I would write test code for this.
Else, I would at least raise a different exception, since the resp.text is in my case empty and the
status code is not returned.

- [ ] New code is annotated.
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on the [KeepAChangeLog] project.
 
 ## Unreleased
 
+### Added
+- [#739] Better error message for providers which return HTTP Error 405 on userinfo
+
+[#739] https://github.com/OpenIDC/pyoidc/pull/739/
+
 ## 1.2.0 [2020-02-05]
 
 ### Fixed
--- a/src/oic/oic/__init__.py
+++ b/src/oic/oic/__init__.py
@@ -28,6 +28,7 @@ from oic import rndstr
 from oic.exception import AccessDenied
 from oic.exception import AuthnToOld
 from oic.exception import AuthzError
+from oic.exception import CommunicationError
 from oic.exception import MissingParameter
 from oic.exception import ParameterError
 from oic.exception import PyoidcError
@@ -938,6 +939,12 @@ class Client(oauth2.Client):
                 )
         elif resp.status_code == 500:
             raise PyoidcError("ERROR: Something went wrong: %s" % resp.text)
+        elif resp.status_code == 405:
+            # Method not allowed error
+            allowed_methods = [x.strip() for x in resp.headers["allow"].split(",")]
+            raise CommunicationError(
+                "Server responded with HTTP Error Code 405", "", allowed_methods
+            )
         elif 400 <= resp.status_code < 500:
             # the response text might be a OIDC message
             try:
--- a/tests/test_oic.py
+++ b/tests/test_oic.py
@@ -12,6 +12,7 @@ from jwkest.jws import left_hash
 from jwkest.jwt import JWT
 from requests import Response
 
+from oic.exception import CommunicationError
 from oic.exception import RegistrationError
 from oic.oauth2.exception import OtherError
 from oic.oauth2.message import SINGLE_OPTIONAL_STRING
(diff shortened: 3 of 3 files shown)


tpazderka:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/tpazderka/pyoidc/pullRequest?prid=4872878&bid=16153260)

tpazderka:
To me this looks like a provider that does not honor the specifications as they explicitly say that the userinfo must support GET and POST.

https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

JustKiddingCode:
> To me this looks like a provider that does not honor the specifications as they explicitly say that the userinfo must support GET and POST.
> 
> https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

Good catch, had only the OAuth spec in mind where it just states that it must return an appropriate error code.
Well for the provider, I think I can trigger the developers to support both, but in the mean time I'd be happy with this fix. Especially, because pyoidc won't give me a usable error.

tpazderka:
I feel that this is a workaround I personally don't like very much. I would be in favor of modifying the current code to give better errors if the `text` is empty.

Another issue I see with this is that it will happily do any method it finds in headers...

JustKiddingCode:
Changed it, so that it raises CommunicationError.
I think this might be the appropriate exception?

tpazderka:
Seems appropriate. Could you please add relevant tests?

JustKiddingCode:
I've written the test code for my pull request, but i think the other error handling messages aren't covered by the current tests (need to look into that).
I will expand my test case, so that the other HTTP status code handling is also covered.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on the [KeepAChangeLog] project.
 
 ## Unreleased
 
+### Added
```
tpazderka: These are extra due to conflict resolution.

src/oic/oic/__init__.py
```diff
@@ -938,6 +939,12 @@ def do_user_info_request(
                 )
         elif resp.status_code == 500:
             raise PyoidcError("ERROR: Something went wrong: %s" % resp.text)
+        elif resp.status_code == 405:
```
tpazderka: Have you thought about updating the general error handling (L948) to better handle this?

It looks to me that adding the status code to the thrown exception on L953, might be enough to put developers on the right track.

JustKiddingCode: Sorry for late reply, had to fix stuff for my study.
The problem with the general exception is that the ALLOW header exists only if a 405 is thrown.
Right now the exception a lib user gets is thrown at L953 (since the request is no valid json). To reach the
general error handling we would filter 405 out of the 400-500 range part.
This seems more complex to me than my solution.

JustKiddingCode:
Took a deeper look into the specs. Is the general error handling used by any known provider?
[1] just links to the OAuth Spec [2] that doesn't suggest to use json here and instead advise to look into the WWW-Authenticate header.

[1] https://openid.net/specs/openid-connect-core-1_0.html#UserInfoError
[2] https://tools.ietf.org/html/rfc6750#section-3

src/oic/oic/__init__.py
tpazderka: I was thinking about simply adding the `status_code` to the error message.

But this works as well.

tpazderka:
I am not sure if that is actually ever used.

The specs say, that the WWW-Authenticate should be used in cases of authentication related changes and in other cases, the respective HTTP status code.

The JSON is not specifically mentioned, but the OIC usually uses that everywhere, so I assume that it should be OK.

The code probably needs a rebase to get the CI to pass, but otherwise looks good.

JustKiddingCode:
> 
> The code probably needs a rebase to get the CI to pass, but otherwise looks good.
Should be ok now.

JustKiddingCode:
Seems that pylama(?) isn't found in the travis ci. Anyway, I think this is not related to my code.

tpazderka:
No, it is and it is complaining about unused varibles: https://travis-ci.org/OpenIDC/pyoidc/jobs/654394698#L294

JustKiddingCode:
> No, it is and it is complaining about unused varibles: https://travis-ci.org/OpenIDC/pyoidc/jobs/654394698#L294

Thanks, i was irritated about the red error message.
hopefully this will fix it :)

PR #740, opened 2020-01-24
Fix: SINGLE_OPTIONAL_BOOLEAN allow to serialise/deserialise as JSON

kkszysiu:
Do the parsing of boolean objects in case it's set to boolean as json string

Prevents error as such:
```
  File "/home/rebelmouse/env/rebelmouse/local/lib/python2.7/site-packages/oic/oic/message.py", line 339, in verify
    self["id_token"] = verify_id_token(self, **kwargs)
  File "/home/rebelmouse/env/rebelmouse/local/lib/python2.7/site-packages/oic/oic/message.py", line 296, in verify_id_token
    idt = IdToken().from_jwt(_jws, **args)
  File "/home/rebelmouse/env/rebelmouse/local/lib/python2.7/site-packages/oic/oauth2/message.py", line 686, in from_jwt
    return self.from_dict(jso)
  File "/home/rebelmouse/env/rebelmouse/local/lib/python2.7/site-packages/oic/oauth2/message.py", line 359, in from_dict
    self._add_value(skey, vtyp, key, val, _deser, null_allowed)
  File "/home/rebelmouse/env/rebelmouse/local/lib/python2.7/site-packages/oic/oauth2/message.py", line 395, in _add_value
    raise ValueError('"{}", wrong type of value for "{}"'.format(val, skey))
ValueError: "true", wrong type of value for "email_verified"
```
---

### DIFF
--- a/src/oic/oic/message.py
+++ b/src/oic/oic/message.py
@@ -98,7 +98,7 @@ def json_rest(val, sformat=None, lev=0):
 
 
 # value type, required, serializer, deserializer, null value allowed
-SINGLE_OPTIONAL_BOOLEAN = ParamDefinition(bool, False, None, None, False)
+SINGLE_OPTIONAL_BOOLEAN = ParamDefinition(bool, False, json_ser, json_deser, False)
 SINGLE_OPTIONAL_JSON_WN = ParamDefinition(dict, False, json_ser, json_deser, True)
 SINGLE_OPTIONAL_JSON_CONV = ParamDefinition(dict, False, json_conv, json_rest, True)
 SINGLE_REQUIRED_INT = ParamDefinition(int, True, None, None, False)


kkszysiu:
Hey @tpazderka , can you help me to verify it it's valid case?
In my case I was getting error shown above. The code change I've made makes sense but I do not know if it's something that my identity provider is implementing against the spec or the library is missing this.

tpazderka:
I think that this is an out of spec behavior. Boolean objects must be represented as JSON booleans.

https://openid.net/specs/openid-connect-core-1_0.html#JSONSerialization

igaitan:
Hi all, i found that using AWS Cognito as OpenID provider got the same error as described here.
I uses properly this patch to make it works properly with this provider. 
I understand clearly that is out of spec, but really does not breaks nothing as the serializer case is the last one used and give more compatibility for the real providers out there.
Anyway thanks for all guys!

lfdominguez:
Hi, great, this happen when you use Apple Sign In, i needed to hack the code to accept a boolean as string...

image

tpazderka:
This patch would break the adherence to specs, so I am not going to accept it.

kkszysiu:
Makes total sense. But it still makes sense to keep it here, just for people who will need it. Our case was also related to Apple Sign In and if big providers such as Amazon or Apple are affected, then it still might be needed solution for some of us.

tpazderka:
It is still searchable even closed, isn't it?

kkszysiu:
Yup. That should be enough :-)

Issue #754, opened 2020-07-16
ClientSecretBasic should not url encode credentials before encoding in base64


gbmarc1:
Bug in the ClientSecretBasic.
My request was working when using PostMan but not when using this library. I then realized that the Authorization header was not the same between the two requests. My client secret had non-friendly url characters.

```
class ClientSecretBasic(ClientAuthnMethod):
   ...
    def construct(self, cis, request_args=None, http_args=None, **kwargs):
        ...
        credentials = "{}:{}".format(quote_plus(user), quote_plus(passwd))  # This line should be "{}:{}".format(user, passwd)
        authz = base64.b64encode(credentials.encode("utf-8")).decode("utf-8")

```

https://github.com/rohe/pyoidc/blob/75275e90457ed7eec71ae383e38f5e43995981c2/src/oic/utils/authn/client.py#L125

schlenk:
Indeed, the `quote_plus` is wrong in that case. 

RFC 7617/RFC5234/RFC7613 place a few minor restrictions on the valid password and user strings, but those are mostly harmless (no CTL chars %x00-1F / %x7F and no : in the username and some ASCII compatibility or unicode NFC normalization.

gbmarc1:
#755

tpazderka:
Looking at the [commit message](https://github.com/rohe/pyoidc/commit/c8b60d6194b5b1bb8a223cacf0e550ca18332a56) message that introduced that, I think that the `quote_plus` might need to stay. Or the next line might need to be changed to `urlsafe_b64encode`.

I am not entirely sure on this one.

schlenk:
I think the commit message is a little misleading. 

When passing the client_secret via HTTP Basic Auth in the header, the quote_plus is probably wrong.

BUT the OAuth2/OpenID connect spec also optionally allows to pass the client secret/id in the body of the request. And that usecase obviously needs the form encoding. (RFC 6749 2.3.1 or OpenID connect section 9) like all parameters in the body of the request.

On the other hand, the client_id/user is more restricted than Basic Auth allows, due to the various usecases where the client ID is passed via URL parameters, but the secret is not.

schlenk:
Ok, misread the spec there. RFC 6749 explicitly mentions the urlencoding in 2.3.1, but HTTP Basic Auth RFC 7617 is less strict.

> The client identifier is encoded using the
> "application/x-www-form-urlencoded" encoding algorithm per
> Appendix B, and the encoded value is used as the username; the client
[...]

PR #755, opened 2020-07-16
Hotfix/#754 client secret basic remove quote plus

gbmarc1:
- [ ] New code is annotated.
---

### DIFF
--- a/src/oic/utils/authn/client.py
+++ b/src/oic/utils/authn/client.py
@@ -121,7 +121,7 @@ class ClientSecretBasic(ClientAuthnMethod):
         if "headers" not in http_args:
             http_args["headers"] = {}
 
-        credentials = "{}:{}".format(quote_plus(user), quote_plus(passwd))
+        credentials = "{}:{}".format(user, passwd)
         authz = base64.b64encode(credentials.encode("utf-8")).decode("utf-8")
         http_args["headers"]["Authorization"] = "Basic {}".format(authz)
 
--- a/tests/utils/test_authn_client.py
+++ b/tests/utils/test_authn_client.py
@@ -7,9 +7,9 @@ from oic.oic.message import AuthorizationRequest
 from oic.utils.authn.client import get_client_id
 
 CDB = {
-    "number5": {"client_secret": "drickyoughurt"},
+    "number5": {"client_secret": "drickyoughurt*"},
     "token_client": {},
-    "expired": {"client_secret": "drickyoughurt", "client_secret_expires_at": 1},
+    "expired": {"client_secret": "drickyoughurt*", "client_secret_expires_at": 1},
     "secret_token": "token_client",
     "expired_token": "expired",
 }
@@ -45,11 +45,11 @@ class TestGetClientID(object):
             get_client_id(self.cdb, AuthorizationRequest(), "mumbo jumbo")
 
     def test_basic_authn_client_ok(self):
-        authn = "Basic " + b64encode(b"number5:drickyoughurt").decode()
+        authn = "Basic " + b64encode(b"number5:drickyoughurt*").decode()
         assert get_client_id(self.cdb, AuthorizationRequest(), authn)
 
     def test_basic_authn_client_missing(self):
-        authn = "Basic " + b64encode(b"missing:drickyoughurt").decode()
+        authn = "Basic " + b64encode(b"missing:drickyoughurt*").decode()
         with pytest.raises(FailedAuthentication):
             get_client_id(self.cdb, AuthorizationRequest(), authn)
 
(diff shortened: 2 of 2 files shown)


tpazderka:
Are you willing to change this to do not use the `quote_plus` based on a kwarg as mentioned in #754 ?

gbmarc1:
Yes I will do. However, would it be more flexible to add an optional argument where you can specify the pre-encoding (before base64) of the user-pass instead of a boolean to quote_plus or not. 
This would enable any encoding as per the point 3 of the RFC7617 spec.

```
   To receive authorization, the client

   1.  obtains the user-id and password from the user,

   2.  constructs the user-pass by concatenating the user-id, a single
       colon (":") character, and the password,

   3.  encodes the user-pass into an octet sequence (see below for a
       discussion of character encoding schemes),

   4.  and obtains the basic-credentials by encoding this octet sequence
       using Base64 ([RFC4648], Section 4) into a sequence of US-ASCII
       characters ([RFC0020]).

   The original definition of this authentication scheme failed to
   specify the character encoding scheme used to convert the user-pass
   into an octet sequence.  In practice, most implementations chose
   either a locale-specific encoding such as ISO-8859-1 ([ISO-8859-1]),
   or UTF-8 ([RFC3629]).  For backwards compatibility reasons, this
   specification continues to leave the default encoding undefined, as
   long as it is compatible with US-ASCII (mapping any US-ASCII
   character to a single octet matching the US-ASCII character code).
```

src/oic/__init__.py
```diff
@@ -24,7 +24,7 @@
         )
 
 __author__ = "Roland Hedberg"
-__version__ = "1.2.0"
+__version__ = "1.3.0"
```
tpazderka, line 27: Please do not update the version. It will be updated before the release in a separate commit.

src/oic/utils/authn/client.py
```diff
@@ -121,8 +121,21 @@ def construct(self, cis, request_args=None, http_args=None, **kwargs):
...
-        authz = base64.b64encode(credentials.encode("utf-8")).decode("utf-8")
+        if "encoding" in kwargs:
+            encoding = kwargs["encoding"]
+        else:
+            encoding = "utf-8"
+
+        if "url_encoded" in kwargs:
+            url_encoded = kwargs["url_encoded"]
+        else:
+            url_encoded = True
```
tpazderka, line 132: I would rather define the kwargs with their default values in the method signature (and also document in the docstring).

tpazderka:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/tpazderka/pyoidc/pullRequest?prid=5865977&bid=19200421)

src/oic/utils/authn/client.py
tpazderka: Hm, it looks like adding the annotation triggers complains from `mypy`. I am not sure if there is an easy way out of here.
Will have a look.

There are some other quality issues that need to be solved.

tpazderka: OK, the mypy complain is because of the `cis` arg that is not in the superclass. One way to fix this is to remove the typing annotations. The other way is to make the signature the same (by making the `cis` a kwarg rather than arg).

I would go with the latter. If you are not willing to do that, make a separate issue and I will deal with that and you can base your PR on top of that.

tpazderka:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/tpazderka/pyoidc/pullRequest?prid=5865977&bid=19200421)

gbmarc1:
> Looks good code-wise.
> 
> There are still two reported errors from mypy that need to be fixed:
> 
> ```
> tests/test_client.py:88: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")
> 303tests/test_client.py:90: error: Argument 1 to "join" of "bytes" has incompatible type "Tuple[str, str]"; expected "Iterable[Union[ByteString, memoryview]]"
> ```

I don't see any problem with the code

tests/test_client.py
```diff
@@ -58,16 +58,40 @@ def client():
...
+            kwargs.update({"url_encoded": url_encoded})
+
+        if encoding is None:
+            encoding = "utf-8"
+        else:
+            kwargs.update({"encoding": encoding})
+
+        http_args = csb.construct(cis, **kwargs)
+
+        user, passwd = user.encode(encoding), passwd.encode(encoding)
```
tpazderka, line 88: This assigns `bytes` to what was preciously `str`, mypy doesn't like that it may cause programming errors.

It should be solved by using a new variable, or encoding directly in L90.

tpazderka: That should actually solve both mypy complains...

PR #779, opened 2021-05-17
Upgrade pip in travis

schlenk:
Try to upgrade pip to get cryptography .whl working.

- [ ] New code is annotated.
---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
+language: python
+dist: bionic
+
+sudo: required
 services:
   docker
-
-python: 3.6
 
 addons:
   hosts:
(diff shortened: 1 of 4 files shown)


schlenk:
Should be mostly fixed now. Some small issues with quality left (bandit complains about shelve and the ruby linkchecker seems to have some issues with changelog), but the cryptography stuff seems to work again.

tpazderka:
Are we going to keep travis for the integration tests? It should be possible to convert to GA as well, but we might keep it as is for now.

schlenk:
I think Travis is pretty much dead in the short term, just wanted to get things working again. GA would surely be the way forward.

PR #782, opened 2021-05-27
doc: fix Docker URL

eliasp:
Replace the broken double-pasted URL with a working one pointing to the
downloads page for Docker CE.


schlenk:
thanks for the fix.

PR #783, opened 2021-06-09
fix: use logger instance rather than raw logging

Titotix:
  - No changes relevant to users
  - No documentation update needed
- [ ] New code is annotated.
  - Not relevant
  - Not relevant
---

Hey,

I replace a debug log line using directly `logging` by the use of the `logger` instance.
Otherwise, while configuring `oic` logging to level `INFO` in my application, I still get a `DEBUG` root login `KeyBundle fetch keys from: <url>`
I can't figure out any reason to use `logging` here.

### DIFF
--- a/src/oic/utils/keyio.py
+++ b/src/oic/utils/keyio.py
@@ -190,7 +190,7 @@ class KeyBundle(object):
             args["headers"] = {"If-None-Match": self.etag}
 
         try:
-            logging.debug("KeyBundle fetch keys from: %s", self.source)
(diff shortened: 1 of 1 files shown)


schlenk:
Thanks for spotting this. Looks like an oversight. yes, this should use logger.

tpazderka:
Thanks, we will need to fix the CI for the updated `mypy` and then we will merge this.

tpazderka (approved):

Issue #789, opened 2021-07-02
Client.parse_response can handle dict's type doesn't allow it.


Hultner:
## Background
I'm using `pyoidc` in a ASGI (starlette-based) application. When using a framework like starlette or based upon it the query param's are automatically decoded into dicts.

For my client callback I followed the documentation reference implementation for a code flow. And when I construct my `AuthorizationResponse` I need to encode my data back into json (or urlencoded) for the library to parse it again via `client.parse_response`, causing an unnecessary extra encode/decode roundtrip. 

## Current behaviour
I read the source code and it became apparent that the code technically supports me to provide a dict directly, this is however disallowed by the types both for `sformat: ENCODINGS` where `ENCODINGS = Literal["json", "urlencoded"]` and `info: str`. 

`Client.parse_response` can technically handle dict's but it's unsupported by its type. This can be discovered by looking further into the implementation of `Message.deserialize`, which uses `func = getattr(self, "from_%s" % method)` where `method` originates from `sformat`. Since a `Message.from_dict` exists there's no technical limitation to supporting it from my point of view. I've tried this and it appears to function properly.

## Example code
Currently supported way of doing it (with unnecessary json roundtrip).
```python
auth_resp = client.parse_response(
    AuthorizationResponse,
    info=json.dumps({"code": code, "state": state, "code": code}),
)
```

Works but isn't allowed by the types, hence relies on implementation details not part of the API.
```python
auth_resp = client.parse_response(
    AuthorizationResponse,
    # Can we do this without json dump?
    # info=json.dumps({"code": code, "state": state, "code": code}),
    # Yes we can but it's not officially supported and relies on implementation details.
    # This could break in future versions so it might be worth the extra json round-trip.
    sformat="dict",
    info={"code": code, "state": state, "code": code},
)
```

## Suggestion
Change the type signature of `Client.parse_response` and officially support input via a dictionary.
 
```diff
class Client(PBase):
    ...

    def parse_response(
        self,
        response: Type[Message],
-       info: str = "",
+       info: Union[str, dict] = "",
        # Or make the change in ENCODINGS directly if it work's everywhere.
-       sformat: ENCODINGS = "json",
+       sformat: Union[ENCODINGS, Literal['dict']] = "json",
        state: str = "",
        **kwargs
    ) -> Message:
```

If this is not suitable for other reasons I completely understand but otherwise this would make a lot of sense from my point of view. I can supply a patch/PR if you're willing to incorporate this change.

[...]

PR #790, opened 2021-07-05
Support for dict in Client.parse_response formats

Hultner:
- [x] New code is annotated.
---
In response to #789

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,14 @@ The format is based on the [KeepAChangeLog] project.
 - [#763] Drop python 3.5 support
 
 ### Added
+- [#790] Support for dict in Client.parse_response formats
 - [#739] Better error message for providers which return HTTP Error 405 on userinfo
 - [#723] Add settings class to handle settings related to Client and Server
 
 [#723]: https://github.com/OpenIDC/pyoidc/pull/723/
 [#739]: https://github.com/OpenIDC/pyoidc/pull/739/
 [#763]: https://github.com/OpenIDC/pyoidc/pull/763/
+[#790]: https://github.com/OpenIDC/pyoidc/pull/790
 
 ## 1.2.0 [2020-02-05]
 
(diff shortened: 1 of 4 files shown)


schlenk (commented):
It seems line 630ff in parse_response() needs some cast() to make mypy happy for the urlencoded path still.

```
        if sformat == "urlencoded":
            str_info = cast(str, info)
            info = self.get_urlinfo(str_info)
```

Looks good otherwise. Could you add this please?

Hultner:
> It seems line 630ff in parse_response() needs some cast() to make mypy happy for the urlencoded path still.
> 
> ```
>         if sformat == "urlencoded":
>             str_info = cast(str, info)
>             info = self.get_urlinfo(str_info)
> ```
> 
> Looks good otherwise. Could you add this please?

Fixed it now. I also added an additional type guard so that an error is raised if info is passed as an dict but sformat isn't a dictionary.

schlenk (approved):
Seems the extra isinstance() check pushes pylama over the threshold of its complexity measure due to the extra if. I just pushed the limit a little higher as it is arbitrary anyway and will refactor it later. 

The Travis check is basically dead (due to travis stopping its free service), so no issue there.

Hultner:
> Seems the extra isinstance() check pushes pylama over the threshold of its complexity measure due to the extra if. I just pushed the limit a little higher as it is arbitrary anyway and will refactor it later.
> 
> The Travis check is basically dead (due to travis stopping its free service), so no issue there.
Thanks for the swift feedback and process :) 

I saw that the last release was last year. Is there any plan for when the next release will be? I've seen some other changes in master which aren't in the released version which I'd like to have.

schlenk:
Yes, a release is planned soonish, i think when @tpazderka finds the time to do one.

Hultner:
Nice, great to hear. Looks like you've made some python 3.9 improvements since the last release.

Issue #800, opened 2022-02-26
Client Credentials Flow request fails because get_request_type always returns AccessTokenRequest for token_endpoint


infohash:
When calling this method `Client.do_access_token_request` of `oic.oauth2` package, Client Credentials Flow request fails due to the error `oic.oauth2.exception.GrantError: No grant found for state:''`. Here's why:

To use client credentials, `do_access_token_request(request_args=request_args, endpoint=self._client.token_endpoint)` method is called.

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/__init__.py#L872-L885

Then this line above returns `AccessTokenRequest` from here:

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/message.py#L1196-L1203

Fetches from the line below:

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/message.py#L1220

Then `AccessTokenRequest` is passed as a positional argument `request`:

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/__init__.py#L895-L904

Which then calls `construct_AccessTokenRequest` from the code below:

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/__init__.py#L547-L562

This line above is the main cause, though the error is not originating from here. It is not explicitly providing the keyword argument `request=CCAccessTokenRequest` which means that `construct_AccessTokenRequest` will try to resolve it itself in the line below:

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/__init__.py#L435-L449

This if statement above shouldn't have been executed if request keyword argument was provided. This if statement then returns `AccessTokenRequest` but we wanted `CCAccessTokenRequest`. Because of this, another if statement becomes true which should have been false. Here:

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/__init__.py#L452-L468

---

# Solution

[...]

PR #801, opened 2022-02-26
Added a new test case for client credentials flow

infohash:
~Addresses the issue with `oic.oauth2.Client.do_access_token_request` for Client Credentials Flow. See #800~


infohash:
With this fix, if the `request` is `CCAccessTokenRequest`, the `getattr(self, "construct_%s" % request.__name__)` will raise `AttributeError` exception which falls back to `self.construct_request(request, request_args, extra_args)`. We can either implement a new method `construct_CCAccessTokenRequest` or let it fallback to `construct_request` method. Which one do you think is better?

https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oauth2/__init__.py#L547-L564

schlenk:
I think i fixed the quality tooling now. Can you please rebase?

infohash:
It's still failing the checks. Can you address it?

schlenk:
Yes. It just fails a check from "black". 

Oh no! 💥 💔 💥
1 file would be reformatted, 108 files would be left unchanged.

infohash:
Reformatted using black. Awaiting for review.

src/oic/oauth2/__init__.py
```diff
@@ -875,14 +880,16 @@ def do_access_token_request(
         state: str = "",
         body_type: ENCODINGS = "json",
         method="POST",
+        request=None,
```
tpazderka, line 883: The same change should be done in `src/oic/oic/__init__.py` implementation of `Client`.

Also a test of this behaviour would be a nice addition.

infohash: Done. Awaiting approval 🙂.

infohash:
It's ready to be merged.

tpazderka:
I am wondering if the `request` argument should actually be there at all. If I am reading this correctly, this can be solved by setting your own `message_factory` with `CCAccessTokenRequest` in the `token_endpoint`. Or am I missing something?

infohash:
It can be done like that, this code was in the test case:

```python
class CCMessageFactory(OauthMessageFactory):
    """We are doing client credentials."""

    token_endpoint = MessageTuple(CCAccessTokenRequest, AccessTokenResponse)

self.client.message_factory = CCMessageFactory
```

We can have this class implemented in the library.

infohash:
@schlenk I have reverted the code but I've addressed the test cases for client credentials which can be merged.

tests/test_oauth2.py
```diff
@@ -420,6 +424,11 @@ def test_request_info_with_req_and_extra_args(self):
         assert h_args == {}
         assert isinstance(cis, AuthorizationRequest)
 
+    def test_request_info_when_request_is_cc_access_token_request(self):
```
tpazderka, line 427: I would like to see some basic verification of the results as well.

tpazderka (changes requested):

tests/test_oauth2.py
infohash: Done. See if this is what you were requesting.

infohash:
@schlenk Can you merge this PR so that test cases in #805 can pass?

infohash:
@tpazderka @schlenk Can you merge this one now? I'll then merge this one in my remaining PRs in the pipeline so they can pass their test cases. I don't want to leave #805 and #806 hanging because there's a draft PR in https://github.com/zamzterz/Flask-pyoidc/pull/132 which depends on them.

I also have a new PR ready which addresses a bug with zero content length response. I'll create it once I clear existing PRs in the pipeline.

tpazderka (approved):

PR #805, opened 2022-03-27
Included id_token_jwt in token response

infohash:
The ID token JWT needs to be passed in the access token response to be usable as id_token_hint for RP-Initiated Logout. Refer to [RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) specification.

The existing code replaces JWT with the verified IdToken instance but it should be preserved. I've added `id_token_jwt` which stores signed and serialised ID token.

PR #806, opened 2022-04-04
Added support for requests session

infohash:
1. `requests.Session` allows session persistence across multiple requests.
2. Custom Transport Adapters
> Transport Adapters provide a mechanism to define interaction methods for an HTTP service. In particular, they allow you to apply per-service configuration.

> Requests ships with a single Transport Adapter, the [HTTPAdapter](https://docs.python-requests.org/en/latest/api/#requests.adapters.HTTPAdapter). This adapter provides the default Requests interaction with HTTP and HTTPS using the powerful [urllib3](https://github.com/urllib3/urllib3) library. Whenever a Requests [Session](https://docs.python-requests.org/en/latest/api/#requests.Session) is initialized, one of these is attached to the [Session](https://docs.python-requests.org/en/latest/api/#requests.Session) object for HTTP, and one for HTTPS.

> Requests enables users to create and use their own Transport Adapters that provide specific functionality. Once created, a Transport Adapter can be mounted to a Session object, along with an indication of which web services it should apply to.

In some enterprise services, there can be a requirement of using custom transport adapters. Allowing a user to directly pass `requests.Session` instance will satisfy such requirements and flexibility to manage and configure http requests

### DIFF
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -11,6 +11,7 @@ from typing import cast
 from urllib.parse import urlparse
 
 from jwkest import b64e
+from requests import Session
 from typing_extensions import Literal
 
 from oic import CC_METHOD
@@ -181,6 +182,7 @@ class Client(PBase):
         timeout=None,
         message_factory: Type[MessageFactory] = OauthMessageFactory,
         settings: PyoidcSettings = None,
+        request_session: Session = None
     ):
         """
         Initialize the instance.
@@ -192,6 +194,8 @@ class Client(PBase):
                  - verify_ssl
                  - client_cert
                  - timeout
+            request_session
+                Instance of :class:`Session` with configuration options.
 
         :param client_id: The client identifier
         :param client_authn_method: Methods that this client can use to
@@ -229,7 +233,7 @@ class Client(PBase):
                 stacklevel=2,
             )
             self.settings.timeout = timeout
-        PBase.__init__(self, keyjar=keyjar, settings=self.settings)
+        PBase.__init__(self, keyjar=keyjar, settings=self.settings, request_session=request_session)
 
         self.sso_db: Optional[SessionBackend] = None
         self.client_id = client_id
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -11,7 +11,6 @@ from typing import cast
 from urllib.parse import urlparse
 
 from jwkest import b64e
-from requests import Session
 from typing_extensions import Literal
 
 from oic import CC_METHOD
@@ -181,8 +180,7 @@ class Client(PBase):
         client_cert=None,
         timeout=None,
         message_factory: Type[MessageFactory] = OauthMessageFactory,
-        settings: PyoidcSettings = None,
-        request_session: Session = None
+        settings: PyoidcSettings = None
     ):
         """
         Initialize the instance.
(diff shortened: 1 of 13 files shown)


src/oic/oauth2/__init__.py
```diff
@@ -181,6 +182,7 @@ def __init__(
         timeout=None,
         message_factory: Type[MessageFactory] = OauthMessageFactory,
         settings: PyoidcSettings = None,
+        request_session: Session = None
```
tpazderka, line 185: I would rather include this in the `settings` class. The number of `kwargs`  is already quite high...

src/oic/oauth2/__init__.py
```diff
@@ -11,6 +11,7 @@
 from urllib.parse import urlparse
 
 from jwkest import b64e
+from requests import Session
```
tpazderka, line 14: If I see correctly, this is going to be used only for the Client code?

tpazderka (changes requested):

src/oic/oauth2/__init__.py
infohash: removed and moved request_session to settings

infohash:
@tpazderka After merging upstream/python-versions, checks here are failing while #805 has passed all checks.

```
SKIPPED:  py37: InterpreterNotFound: python3.7
  py38: commands succeeded
SKIPPED:  py39: InterpreterNotFound: python3.9
SKIPPED:  py310: InterpreterNotFound: python3.10
  docs: commands succeeded
ERROR:   quality: commands failed
Error: Process completed with exit code 1.
```

tpazderka:
You have errors in typing, you can run `tox -equality` to see that.

infohash:
Fixed and ready to be merged. When will be the expected date of release?

src/oic/oauth2/base.py
```diff
@@ -128,7 +128,10 @@ def http_request(self, url, method="GET", **kwargs):
             _kwargs = self.req_callback(method, url, **_kwargs)
 
         try:
-            r = requests.request(method, url, **_kwargs)  # type: ignore
+            if self.settings.request_session:
```
tpazderka, line 131:
```suggestion
            if self.settings.request_session is not None:
```
Testing for `None` is better explicit.

src/oic/utils/settings.py
```diff
@@ -36,6 +38,8 @@ class PyoidcSettings:
             Timeout for requests library.
             Can be specified either as a single float or as a tuple of floats.
             For more details, refer to ``requests`` documentation.
+        request_session
+                Instance of :class:`Session` with configuration options.
```
tpazderka, line 42: As this is client specific, I would rather move it to `ClientSettings`.

tpazderka (changes requested):

src/oic/oauth2/base.py
infohash: Replaced with `if getattr(self.settings, "request_session", None) is not None:`

tpazderka: Hm, I suppose this is due to the move to `ClientSettings` right? I am wondering what is better, whether to limit this to client only or allow server to have this config as well.

But I am leaning towards keeping that as a Client only. Will think about it for a day and we will see.

infohash: > I suppose this is due to the move to ClientSettings right?

Yes.

> I am wondering what is better, whether to limit this to client only or allow server to have this config as well.

In most cases, server only responds to the client requests so http session persistence is mostly needed by the client. The [Advanced Usage](https://docs.python-requests.org/en/latest/user/advanced/#advanced-usage) of the `requests.Session` also explains it from the perspective of the client.

Besides if the user still wants to use session persistence for the provider, it can be done like this.

```python
from oic.utils.settings import ServerSettings

server_settings = ServerSettings()
session = requests.Session()
session.verify = True
server_settings.request_session = session
```

tpazderka (approved):

PR #812, opened 2022-05-06
Fixed parsing for zero content length response

infohash:
The `content-type` in the header is optional when the payload has zero content length. See, [Should Content-Type header be present when the message body is empty?](https://stackoverflow.com/questions/29784398/should-content-type-header-be-present-when-the-message-body-is-empty).

This happens when the request to IdP is made for revoking tokens. In that case, IdP returns a `200 OK` status code with no body and no  `Content-Type` in the response header.

See, [About the revoke request](https://developer.okta.com/docs/guides/revoke-tokens/main/#about-the-revoke-request).

> Revoking a token that is invalid, expired, or already revoked returns a 200 OK status code to prevent any information leaks.

To parse responses with zero content length, just check if the value of `content-length` is 0 and return the status code if it does.

### DIFF
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -747,7 +747,10 @@ class Client(PBase):
             )
 
         if response:
-            if body_type == "txt":
+            if body_type is None:
+                # There is no content-type for zero content length. Return the status code.
+                return reqresp.status_code
+            elif body_type == "txt":
                 # no meaning trying to parse unstructured text
                 return reqresp.text
             return self.parse_response(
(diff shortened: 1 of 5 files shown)


tpazderka (changes requested):
The new code should be covered by tests.

tpazderka (changes requested):
The tests should be unittests, so you should write a test for `parse_request_response` and `verify_header` functions that cover the changes introduced here.

infohash:
@tpazderka Awaiting review.

tpazderka (approved):

infohash:
When the new version will be released?

Issue #551, opened 2018-06-20
PKCE support?


bmerry:
This is more a question than a bug report. Does pyoidc support PKCE (RFC 7636), and if so, how does one use it? I went digging into the code and found [this](https://github.com/OpenIDC/pyoidc/blob/69a7ef59c1f40c29313354e034bc15bfff92921b/src/oic/oic/__init__.py#L660), which seems to handle adding the code challenge to the authorization request, but the code verifier doesn't seem to be saved anywhere to send with the access token request.

tpazderka:
Yes, it looks like a partial implenmentation. The storing of the code_verifier is not handled in the code.

Goldziher:
I assume this issue is stale? If so I will work to add a PR for this in the near future.

schlenk:
A PR would be welcome.

bmerry:
We ended up not using OpenID, so this is no longer relevant to me and I don't mind whether it is kept open for a future PR or closed due to my question being answered.

tpazderka:
I don't mind keeping this open as a TODO.

Goldziher:
I'm afraid our client requires this, so I will implement this and add a PR.

Goldziher:
Well, project cancelled. No go.

PR #815, opened 2022-05-30
"code_challenge_methods_supported" added (https://datatracker.ietf.org/doc/html/rfc8414#section-2)

smalihaider-daasi:
…g/doc/html/rfc8414#section-2)

- [ ] New code is annotated.
---

### DIFF
--- a/src/oic/oic/message.py
+++ b/src/oic/oic/message.py
@@ -14,6 +14,7 @@ from jwkest.jwe import JWEException
 from jwkest.jwe import factory as JWE_factory
 from jwkest.jwt import JWT
 
+from oic import CC_METHOD
 from oic.exception import InvalidRequest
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ The format is based on the [KeepAChangeLog] project.
 
 ## 1.4.0 [2022-05-23]
 
+### Added
```
schlenk, line 12: 1.4.0 is released, open a new section above for the next version please.

schlenk:
The change itself is fine, but the output would be a lie, as the necessary PKCE functionality isn't working yet.
See https://github.com/OpenIDC/pyoidc/issues/551

So unless we get PKCE working first, the list should actually be empty.

PR #816, opened 2022-06-04
Fixed KeyError In Client Registration Info

infohash:
When registration info has redirect_uri missing, it raises `KeyError` exception when method `store_registration_info` of class `oic.extension.Client` is called. Use `get` to access value so that by default it returns `None` if the key is missing.

### DIFF
--- a/src/oic/extension/client.py
+++ b/src/oic/extension/client.py
@@ -340,9 +340,9 @@ class Client(oauth2.Client):
 
     def store_registration_info(self, reginfo):
         self.registration_response = reginfo
-        self.client_secret = reginfo["client_secret"]
-        self.client_id = reginfo["client_id"]
-        self.redirect_uris = reginfo["redirect_uris"]
+        self.client_secret = reginfo.get("client_secret")
+        self.client_id = reginfo.get("client_id")
+        self.redirect_uris = reginfo.get("redirect_uris")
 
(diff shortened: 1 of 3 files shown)


src/oic/extension/client.py
```diff
@@ -340,9 +340,9 @@ def do_authorization_request(
 
     def store_registration_info(self, reginfo):
         self.registration_response = reginfo
-        self.client_secret = reginfo["client_secret"]
-        self.client_id = reginfo["client_id"]
-        self.redirect_uris = reginfo["redirect_uris"]
+        self.client_secret = reginfo.get("client_secret")
+        self.client_id = reginfo.get("client_id")
+        self.redirect_uris = reginfo.get("redirect_uris")
```
schlenk, line 345: This does clash with the definition of self.redirect_uris as a List[str]:

https://github.com/infohash/pyoidc/blob/b6da93d73001981e6afc5bf2e77f47e8e7889b44/src/oic/oauth2/__init__.py#L244

schlenk (commented):
What is the usecase for this change?

RegistrationResponse sets most fields to "required", see:
https://github.com/infohash/pyoidc/blob/b6da93d73001981e6afc5bf2e77f47e8e7889b44/src/oic/oic/message.py#L666

So the only field thats really optional in that place is "client_secret".

PR #823, opened 2022-07-25
move `beaker` to an optional/extras requirement under 'examples'

biyeun:
There is a [CVE](https://github.com/advisories/GHSA-3cwm-7jmm-774w) out for `beaker` with no option to upgrade to a fixed version as the project [appears largely unmaintained](https://github.com/bbangert/beaker).

It appears that `beaker` is only used in the `examples` and is therefore not a necessary requirement to use `oic`. This change moves `beaker` to `extras_requires` in `setup.py` under `examples`, so that this package is not unnecessarily installed as a required sub-dependency on projects listing `oic` as a dependency.


schlenk:
The CVE is kind of bogus, as one can use a json_serializer to enforce json for storage instead of pickle, so it is not unmitigated.
But yes, beaker looks pretty unmaintained and has other issues that could be seen as security hazards (e.g. contradictory documentation how to actually invalidate/delete a session after logout.). 

But i agree that beaker is only needed for examples, so should not be needed for the base package.

PR #828, opened 2022-11-01
Added support for python 3.11

### LINKED ISSUE #827: type hints issue - python 3.11 compatibility
While trying to use the library with python 3.11 we identified at least one issue caused by type hints checks performed in the `__setattr__` method of `utils/settings.py`. 

We get the following error when client_cert is None (default value). 

> oic.utils.settings.SettingsException: client_cert has a type of <class 'NoneType'>, expected any of (<class 'str'>, <class 'tuple'>).

Not tested but setting the type hints as Optional might fix it.

tpazderka:
Close #827

- [ ] New code is annotated.
---

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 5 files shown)


schlenk (approved):
LGTM.

PR #831, opened 2022-11-11
fix for some provider return birthdate:null in userInfo

### LINKED ISSUE #830: birthdate miss lead a exception
Hi, im using pyoidc with [authing](https://www.authing.cn/)，if the app request scope included birthdate, and the provider do not reply a birthdate, it will go a exception as the `verify` do not check if it's a NoneType
```
  def verify(self, **kwargs):
        super().verify(**kwargs)

        if "birthdate" in self:
            # Either YYYY-MM-DD or just YYYY or 0000-MM-DD
            try:
                time.strptime(self["birthdate"], "%Y-%m-%d")
            except ValueError:
                try:
                    time.strptime(self["birthdate"], "%Y")
                except ValueError:
                    try:
                        time.strptime(self["birthdate"], "0000-%m-%d")
                    except ValueError:
                        raise VerificationError("Birthdate format error", self)
```

the userinfo returned by provider is something like this
```
'userinfo': {'name': None, 'nickname': None, 'birthdate': None ...}
```
actually i do not care for the `birthdate`, is there any work-around ?

pkking:
Signed-off-by: Li Chaoran <pkwarcraft@gmail.com>

- [ ] New code is annotated.
---
the code is straightforward, no need any comments :)
fix #830

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on the [KeepAChangeLog] project.
 
 ### Changed
 - [#827] Added support for python 3.11
+- [#830], [#831] Allow null and empty values in UserInfo responses, but filter them out. 
 
 ### Fixed
 - [#826], [#829] Fixed RP-Initiated Logout To Accept id_token_hint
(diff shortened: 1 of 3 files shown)


pkking:
@schlenk  seems the code check failed has nothing to do with this PR?

schlenk:
yes, that is some mypy changes fallout. I'll take a look later today.

schlenk:
I change the fix a bit. Your fix was fine, but would cause an exception when used in a client to fetch userinfo, as the `verify()` returned `false` due to the empty values. Now the code filters out empty or null values from responses.

schlenk (approved):

PR #843, opened 2023-01-30
Fix typing for Cryptodomex

tpazderka:
- [ ] New code is annotated.
---

### DIFF
--- a/src/oic/utils/aes.py
+++ b/src/oic/utils/aes.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
+import sys
 from base64 import b64decode
 from base64 import b64encode
+from typing import Dict
 from typing import Union
 from typing import cast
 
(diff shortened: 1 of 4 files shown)


src/oic/utils/aes.py
```diff
@@ -16,7 +17,11 @@
 
 __author__ = "rolandh"
 
-POSTFIX_MODE = {"cbc": AES.MODE_CBC, "cfb": AES.MODE_CFB, "ecb": AES.MODE_CFB}
+POSTFIX_MODE: dict[str, Union[Literal[2], Literal[3]]] = {
```
schlenk, line 20: seems it should be typing.Dict[] to work, according to the test output.

schlenk (changes requested):
Looks good besides the small glitches with dict vs. typing.Dict

src/oic/utils/aes.py
tpazderka: Yes, that is python 3.9 only :/ Fixed.

schlenk (approved):
LGTM

PR #844, opened 2023-02-10
Don't throw exception for token error response.

zamzterz:
---

When upgrading the `pyoidc` dependency from `1.4.0` to `1.5.0` in [Flask-pyoidc](https://github.com/zamzterz/Flask-pyoidc) it revealed a breaking API change (by breaking [this test](https://github.com/zamzterz/Flask-pyoidc/blob/6c55fbe304a39902a9c6374fd560845218461382/tests/test_pyoidc_facade.py#L205)): A`HttpError` exception is thrown for token error responses, without any possibility of recovering the error response.

This change brings back the [behavior in `1.4.0`](https://github.com/OpenIDC/pyoidc/blob/1.4.0/src/oic/oauth2/__init__.py#L738-L741) which returns the token error response allowing for further handling.

### DIFF
--- a/src/oic/oauth2/__init__.py
+++ b/src/oic/oauth2/__init__.py
@@ -743,14 +743,7 @@ class Client(PBase):
             logger.error("(%d) %s" % (reqresp.status_code, sanitize(reqresp.text)))
             raise ParseError("ERROR: Something went wrong: %s" % reqresp.text)
 
-        if reqresp.status_code in SUCCESSFUL:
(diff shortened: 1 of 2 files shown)


schlenk (approved):
LGTM.

Seems i missed the late conversion to the ErrorResponse type in the `parse_response`() code that runs only after this block, so the issubclass() failed for this case.

zamzterz:
@schlenk Thanks for the quick response! Any chance of getting this in a patch release?

PR #862, opened 2023-07-02
Moved dependency from pydantic to pydantic-settings

infohash:
`BaseSettings` have been moved to `pydantic-settings` package under the same maintainer. As Pyoidc is using `BaseSettings` only in `oic.utils.settings`, we can change the dependency without any problem.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup(
     install_requires=[
         "requests",
         "pycryptodomex",
-        "pydantic",
+        "pydantic-settings",
         "pyjwkest>=1.3.6",
         "mako",
         "cryptography",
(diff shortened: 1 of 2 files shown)


schlenk:
Seems autodoc-pydantic isn't fit for the new v2 api yet either:
https://github.com/mansenfranzen/autodoc_pydantic/issues/146

Your change is okay, so I'm merging this, to have working code. 
The docs will hopefully fix themselves once autodoc_pydantic gets updated.

mansenfranzen:
@infohash @schlenk  You can now test the new **autodoc_pydantic** version supporting pydantic v2 via `pip install autodoc_pydantic==2.0.0`. The dev docs already showcase v2 support (see [here](https://autodoc-pydantic.readthedocs.io/en/main-2.x/index.html)). Migration guide is available [here](https://autodoc-pydantic.readthedocs.io/en/main-2.x/users/faq.html#migration-guide-from-v1-to-v2).

Any feedback is more than appreciated. The current **autodoc_pydantic** version `2.0.0` on pypi remains yanked for one week until final release.

schlenk:
Thanks for the fix @mansenfranzen . I tried it in a branch (https://github.com/CZ-NIC/pyoidc/pull/863) and it seems to fix the issue with the documentation.

PR #869, opened 2024-02-22
Remove a test component which involves testing the daylight saving si…

shearichard-nec:
…tuation for a date in February. This test fails when the test is being run in those parts of southern hemisphere which use daylight saving.

- [x ] The documentation has been updated, if necessary.
- [x] New code is annotated.
---


tpazderka (changes requested):
Can you please rebase on current master? The CI should now pass.

PR #881, opened 2024-07-22
Fix mypy typing

### LINKED ISSUE #880: Fix mypy
New mypy complains about incorrect overrides. It looks to me that some of those are false positives, but we should fix the failing CI anyway...

tpazderka:
- [ ] New code is annotated.
---

Close #880

### DIFF
--- a/src/oic/extension/proof_of_possesion.py
+++ b/src/oic/extension/proof_of_possesion.py
@@ -32,7 +32,7 @@ class PoPProvider(Provider):
         # mapping from signed pop token to access token in db
         self.access_tokens: Dict[JWS, str] = {}
 
-    def token_endpoint(self, dtype="urlencoded", **kwargs):
+    def token_endpoint(self, request="", authn="", dtype="urlencoded", **kwargs):
         atr = AccessTokenRequest().deserialize(kwargs["request"], dtype)
         resp = super(PoPProvider, self).token_endpoint(**kwargs)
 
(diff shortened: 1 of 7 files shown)


src/oic/extension/proof_of_possesion.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, *args, **kwargs):
         # mapping from signed pop token to access token in db
         self.access_tokens: Dict[JWS, str] = {}
 
-    def token_endpoint(self, dtype="urlencoded", **kwargs):
+    def token_endpoint(self, request="", authn="", dtype="urlencoded", **kwargs):
```
schlenk, line 35: Doesn't this technically change the signature, if someone calls this with *args?

It might be better to use token_endpoint(self, dtype="urlencoded", request="", authn="", **kwargs) or mark things as keyword only arguments altogether?

schlenk (approved):
Looks okay. I do not think the minor signature changes should cause any issues here.

src/oic/extension/proof_of_possesion.py
tpazderka: True, will try to reorder that...

tpazderka: Well, no... mypy complains on the "wrong" order, so we have to keep it the way it is here.

I do hope that this piece of code is not widely used and we can bump the version to be incompatible...

schlenk: ok.

PR #883, opened 2024-10-18
Update supported python versions

tpazderka:
- [ ] New code is annotated.
---
3.8 has reached EOL, added new versions.


tpazderka:
@schlenk It looks like `pylama` is not up to speed with development.

I would argue to drop it in favor of `ruff` (and switch to it as a formatter as well to get rid of duality of tools).

schlenk:
Agreed on switching to ruff, it seems to be the better choice currently.

oidc_example/op1/oc_server.py
```diff
@@ -1,52 +1,44 @@
...
-from exceptions import Exception
-from exceptions import IndexError
-from exceptions import KeyboardInterrupt
-from exceptions import KeyError
-from exceptions import OSError
 from logging.handlers import BufferingHandler
-from urlparse import parse_qs
 
+from exceptions import AttributeError, Exception, IndexError, KeyboardInterrupt, KeyError, OSError
 from mako.lookup import TemplateLookup
```
schlenk, line 11: This looks a little surprising from a isort/pep8 styleguide perspective. 
mako.lookup should be third party and not lined up with the stdlib stuff.

But if ruff thinks its the way, I'm fine with it.

schlenk (approved):
LGTM

PR #884, opened 2024-11-06
Fix device flow authorization endpoint

bacek:
Device authorization endpoint is `device_authorization_endpoint` as per https://www.rfc-editor.org/rfc/rfc8628.html#section-4

- [ ] New code is annotated.
---

### DIFF
--- a/src/oic/extension/device_flow.py
+++ b/src/oic/extension/device_flow.py
@@ -89,7 +89,11 @@ class DeviceFlowClient(SingleClient):
         if scope:
             req["scope"] = scope
 
-        http_response = self.host.http_request(self.host.provider_info["device_endpoint"], "POST", req.to_urlencoded())
+        http_response = self.host.http_request(
+            self.host.provider_info["device_authorization_endpoint"],
+            "POST",
+            req.to_urlencoded(),
+        )
 
         response = self.host.parse_request_response(AuthorizationResponse, http_response, "json")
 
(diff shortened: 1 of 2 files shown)


src/oic/extension/device_flow.py
```diff
@@ -94,7 +94,7 @@ def authorization_request(self, scope=""):
             req["scope"] = scope
 
         http_response = self.host.http_request(
-            self.host.provider_info["device_endpoint"], "POST", req.to_urlencoded()
+            self.host.provider_info["device_authorization_endpoint"], "POST", req.to_urlencoded()
```
tpazderka, line 97: Could you please update the formatting as suggested by the formater?

Running `black src/ tests/` should be sufficient.

tpazderka (changes requested):

src/oic/extension/device_flow.py
bacek: > Could you please update the formatting as suggested by the formater?
> 
> Running `black src/ tests/` should be sufficient.

Sorry, totally missed it.

I run `black` on src. Running on `tests` brings too many unrelated changes.

tpazderka: Yeah, sorry... we switched to ruff in the meantime...

bacek: No worries. Updated with ruff.

tpazderka (approved):

PR #889, opened 2025-05-30
Fix quality complaints

tpazderka:
- [ ] New code is annotated.
---

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
     include_package_data=True,
     classifiers=[
         "Development Status :: 4 - Beta",
-        "License :: OSI Approved :: Apache Software License",
+        "Apache 2.0",
         "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -52,7 +52,7 @@
     include_package_data=True,
     classifiers=[
         "Development Status :: 4 - Beta",
-        "License :: OSI Approved :: Apache Software License",
+        "Apache 2.0",
```
schlenk, line 55: If we want to fix the license declarations, we should probably upgrade to the SPDX License Expressions (https://spdx.org/licenses/Apache-2.0.html) as also done in https://peps.python.org/pep-0639/, which basically deprecates this usage in classifiers: https://peps.python.org/pep-0639/#deprecate-license-classifiers

schlenk (changes requested):
We should switch to PEP 639 license declarations.

